PR #1, opened 2021-10-01
Update config.py

jackpertschuk:


pinecone/config.py
```diff
@@ -137,8 +137,8 @@ def _preprocess_and_validate_config(self, config: dict) -> dict:
         and raises ValueError in case an invalid value was specified.
         """
         # general preprocessing and filtering
-        result = {k.strip(): v.strip() for k, v in config.items() if v is not None}
-        result = {k: v for k, v in result.items() if k in ConfigBase._fields}
+#         result = {k.strip(): v.strip() for k, v in config.items() if v is not None}
```
benjaminran, line 140: Why comment out this line?

jackpertschuk: it throws an error, because v for openapi_config is of type OpenAPI "Configuration", which has no `strip()` method

jackpertschuk: we can fix this however you want

benjaminran: Ah, makes sense

benjaminran (approved):

PR #10, opened 2021-11-08
Creating grpc interace for vector service

rajat08:


pinecone/core/grpc/index_grpc.py
```diff
@@ -155,42 +160,160 @@ def wrapped():
...
+
+    def __repr__(self):
+        return str({'vectors': self.vectors, 'namespace': self.namespace})
+
+
+def parse_fetch_response(response: dict):
+    vd = {}
+    vectors = response.get('vectors')
+    if not vectors:
+        return None
```
benjaminran, line 193: This might be breaking compatibility with the openapi client - maybe there it returns an object with an empty vectors field?

pinecone/core/grpc/index_grpc.py
```diff
@@ -155,42 +160,160 @@ def wrapped():
...
-               request: 'vector_service_pb2.UpsertRequest',
-               timeout: int = None,
-               metadata: Dict[str, str] = None):
-        return self._wrap_grpc_call(self.stub.Upsert, request, timeout=timeout, metadata=metadata)
+    @sentry
+    @validate_and_convert_errors
+    def upsert(self, vectors, **kwargs):
+        def _vector_transform(item):
+            if isinstance(item, Vector):
+                item.metadata
```
benjaminran, line 255: typo line

pinecone/core/grpc/index_grpc.py
```diff
@@ -155,42 +160,160 @@ def wrapped():
...
+class QueryResult(object):
+    def __init__(self, matches: list, namespace: str):
+        self.matches = matches
+        self.namespace = namespace
+
+    def __repr__(self):
+        return str({'matches': self.matches, 'namespace': self.namespace})
+
+
+class QueryResponse(object):
```
benjaminran, line 222: The openapi client model objects allows accessing fields like:
- resp.get('results', default_value) 
- resp['results']
- resp.results

I guess what you have is just the resp.results style? Maybe that's fine as long as we don't show usage of the other styles anywhere in our docs (I think we don't, but could you double check in core/api-docs/code_samples?

Also keep in mind adding pydoc for these classes - that's something that we'll want soon, but doesn't have to be this PR.

rajat08:
@benjaminran  can you review again?

pinecone/core/grpc/index_grpc.py
```diff
@@ -4,20 +4,30 @@
...
+from pinecone.core.utils import _generate_request_id, dict_to_proto_struct, fix_tuple_length, proto_struct_to_dict
 from pinecone.core.utils.sentry import sentry_decorator as sentry
 from pinecone.core.grpc.protos.vector_service_pb2_grpc import VectorServiceStub
 from pinecone.core.grpc.retry import RetryOnRpcErrorClientInterceptor, RetryConfig
 from pinecone.core.utils.constants import MAX_MSG_SIZE, REQUEST_ID, CLIENT_VERSION
+from pinecone.core.grpc.protos.vector_service_pb2 import Vector as GRPCVector, QueryVector as GRPCQueryVector, \
+    UpsertRequest, DeleteRequest, \
+    QueryRequest, FetchRequest, DescribeIndexStatsRequest
+from pinecone.core.utils.error_handling import validate_and_convert_errors
+from pinecone import Vector as _Vector
```
benjaminran, line 25: Can you test doing an `import python` with this? This might be a circular import, where this module tries to import a symbol from the pinecone module while that module is still half-initialized and trying to import from here. If so, import pinecone will fail.

You should be able to just use the pinecone.core.client.models package instead.

pinecone/core/grpc/index_grpc.py
```diff
@@ -155,42 +153,128 @@ def wrapped():
...
+            raise ValueError(f"Invalid query vector value passed: cannot interpret type {type(item)}")
+
+        _QUERY_ARGS = ['namespace', 'top_k', 'filter', 'include_values', 'include_metadata']
+        if 'filter' in kwargs:
+            kwargs['filter'] = dict_to_proto_struct(kwargs['filter'])
+        request = QueryRequest(queries=list(map(_query_transform, queries)),
+                               **{k: v for k, v in kwargs.items() if k in _QUERY_ARGS})
+
+        response = self._wrap_grpc_call(self.stub.Query, request, timeout=timeout)
+        json_response = json_format.MessageToDict(response)
```
benjaminran, line 267: This may be hurting performance a lot to convert from protobuf to json just to then convert to the response model class. Might defeat the purpose of grpc/protobuf somewhat. Something to keep in mind - anyways we'll see when we performance test the grpc client vs the json client.

benjaminran (approved):

pinecone/core/grpc/index_grpc.py
rajat08: Yeah I have been wondering what the impact for this would be. But to get the responses to be human-readable we need to spend some effort in conversions. I will do a quick comparison to see if things are any different

pinecone/core/grpc/index_grpc.py
mutayroei: @rajat08 this PR was merged although this line is still here

pinecone/core/utils/__init__.py
```diff
@@ -70,6 +72,37 @@ def get_user_agent():
     return user_agent
 
 
+def dict_to_proto_struct(d: dict) -> Struct:
```
mutayroei, line 75: @rajat08 This definitions breaks the import since Struct is not defined for non-grpc client

PR #13, opened 2021-11-14
Handle ImportError gracefully for non-grpc clients

mutayroei:
[ch9855]


mutayroei:
@benjaminran We can do it either here or inside the gRPC index class. WDYT?

benjaminran (approved):

PR #14, opened 2021-11-25
Fix exception handling

rajat08:


mutayroei:
[ch10321]

pinecone/core/grpc/index_grpc.py
```diff
@@ -67,7 +67,10 @@ class GRPCIndexBase(ABC):
     Base class for grpc-based interaction with Pinecone indexes
     """
 
-    def __init__(self, name: str, channel=None, grpc_config: GRPCClientConfig = None, _endpoint_override: str = None):
+    _pool = None
+
+    def __init__(self, name: str, channel=None, grpc_config: GRPCClientConfig = None, _endpoint_override: str = None,
```
mutayroei, line 72: @rajat08  I believe that for the default index the parameter is called index_name

mutayroei:
@benjaminran  can you review?

rajat08:
@mutayroei @benjaminran 
Since all the info form the grpc error is condensed in the debug string, we can pass this to PineconeException and get a 'nicer sounding' error. However _InactiveRPCError would be visible in the trace. Would the ideal behavior be to raise PineconeError at an earlier step?

benjaminran (changes requested):
Posted a couple comments about the async logic.

The error handling logic looks good! I think it's a good thing to include the _InactiveRpcError deeper inside the trace.

mutayroei (approved):

PR #16, opened 2021-12-15
[WIP]Using gRPC inbuilt async

rajat08:


pinecone/core/grpc/index_grpc.py
```diff
@@ -183,10 +183,11 @@ def parse_query_response(response: dict):
     for match in response['results']:
         namespace = match.get('namespace', '')
         m = []
-        for item in match['matches']:
-            sc = ScoredVector(id=item['id'], score=item.get('score', 0.0), values=item.get('values', []),
-                              metadata=item.get('metadata', {}))
-            m.append(sc)
+        if 'matches' in match:
```
benjaminran, line 186: Can you add unit tests for this function? Ideally for all our grpc response parsing functions here. Should help us catch future such issues sooner.

pinecone/core/grpc/index_grpc.py
```diff
@@ -224,16 +225,19 @@ def _vector_transform(item):
 
         request = UpsertRequest(vectors=list(map(_vector_transform, vectors)), **kwargs)
         timeout = kwargs.pop('timeout', None)
-
-        response = self._wrap_grpc_call(self.stub.Upsert, request, timeout=timeout)
-        return parse_upsert_response(response)
+        if async_req:
+            return self._wrap_grpc_call(self.stub.Upsert.future, request, timeout=timeout)
+        else:
+            return self._wrap_grpc_call(self.stub.Upsert, request, timeout=timeout)
```
benjaminran, line 231: I don't think the exception handling logic in _wrap_grpc_call will still apply correctly in the async case. Maybe we need to wrap the grpc future in a future of our own which delegates to the grpc one but does error handling when results are retrieved.

You could technically even add a "no 3p calls" integ test (creating a local VectorService grpc server as in https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests/unit/_metadata_test.py#L178-L189) to test timeout behaviors and other grpc client things, but only if you think it's worth the effort. Maybe we should just rely on our core suite of integ tests for that, at least for now.

rajat08:
Will add a seperate PR for future error handling

PR #17, opened 2021-12-20
Error handling for future

rajat08:


pinecone/core/grpc/index_grpc.py
```diff
@@ -226,7 +252,8 @@ def _vector_transform(item):
         request = UpsertRequest(vectors=list(map(_vector_transform, vectors)), **kwargs)
         timeout = kwargs.pop('timeout', None)
         if async_req:
-            return self._wrap_grpc_call(self.stub.Upsert.future, request, timeout=timeout)
+            future = self._wrap_grpc_call(self.stub.Upsert.future, request, timeout=timeout)
+            return PineconeGrpcFuture(future)
```
benjaminran, line 256: You should do the same for the delete operation below.

pinecone/core/grpc/index_grpc.py
```diff
@@ -207,6 +207,32 @@ def parse_stats_response(response: dict):
     return DescribeIndexStatsResponse(namespaces=namespace_summaries, dimension=dimension, _check_type=False)
 
 
+class PineconeGrpcFuture:
```
benjaminran, line 210: Looks like you're missing the exception() and traceback() methods from the interface - https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/framework/foundation/future.py#L163

pinecone/core/grpc/index_grpc.py
```diff
@@ -207,6 +207,38 @@ def parse_stats_response(response: dict):
...
+        return self._delegate.add_done_callback(fun)
+
+    def result(self, timeout=None):
+        try:
+            self._delegate.result(timeout=timeout)
+        except _MultiThreadedRendezvous as e:
+            raise PineconeException(e._state.debug_error_string) from e
+
+    def exception(self,timeout=None):
+        with _MultiThreadedRendezvous as e:
```
benjaminran, line 236: You probably meant to use self._delegate.exception() - this drops the original exception as is

PR #34, opened 2022-02-18
Remove sentry tracking by default

rajat08:


mutayroei:
@rajat08 can we close this one?

PR #35, opened 2022-02-18
Remove Sentry Deps

tonyf:


tonyf:
looks like our code formatter decided to re-format some things so this looks like a chunkier pr than it actually is

PR #39, opened 2022-03-03
rename parameter to indexFullness

yarden-slon:


mutayroei:
Does that happen only in the gRPC interface?

mutayroei (approved):

PR #44, opened 2022-03-21
Add async support for queries and fetch

rajat08:


rajat08:
I think this is still possibly not the cleanest code here, I think 'chaining' futures in the pool and copying states is a relatively cleaner, add_done_callback returns None only hence I don't think we can return that to the user directly.
Lmk if you have a better idea

jackpertschuk (changes requested):
I would go for an actual "asyncio" API with "awaits" etc.. instead of these callbacks.

mutayroei:
> I would go for an actual "asyncio" API with "awaits" etc.. instead of these callbacks.

Using the classic async-await will require to change our entire client to use async-await style, won't it? And I don't think it's food idea since most users are not familiar with thay syntax...

mutayroei (approved):

jackpertschuk:
> > I would go for an actual "asyncio" API with "awaits" etc.. instead of these callbacks.
> 
> Using the classic async-await will require to change our entire client to use async-await style, won't it? And I don't think it's food idea since most users are not familiar with thay syntax...

We should add a different class for the async client, no need to change the existing one for users who want the standard client.

mutayroei:
> > > I would go for an actual "asyncio" API with "awaits" etc.. instead of these callbacks.
> > 
> > 
> > Using the classic async-await will require to change our entire client to use async-await style, won't it? And I don't think it's food idea since most users are not familiar with thay syntax...
> 
> We should add a different class for the async client, no need to change the existing one for users who want the standard client.

That would require adding much more code to the client. We can do that if we want, but why not adding that code that Rajat suggested anyway? We already have that in the standard client and in other APIs in the gRPC code. It'll give a good solution to the customers that are waiting for that support. And we can consider having a full asyncio client seperately

mutayroei:
@rajat08 how does the chaning behave in case of errors? Can you test? Does it affect the error messages presented to customers?

pinecone/core/grpc/index_grpc.py
```diff
@@ -77,6 +80,7 @@ def __init__(self, index_name: str, channel=None, grpc_config: GRPCClientConfig
             "service-name": index_name,
             "client-version": CLIENT_VERSION
         }
+        self.pool = ThreadPoolExecutor()
```
benjaminran, line 83: It'd be nice to avoid managing another threadpool if we can. WDYT about reusing PineconeGrpcFuture like upsert/delete already do and doing the response parsing in the result() method? It's simpler, though it may hurt performance. But I'm not actually sure about the performance impact; if response parsing is quick enough relative to the network/backend latency maybe the extra thread switching will hurt more than it helps. It'd be good to get some rough performance numbers on that (one calling thread, response parsing on caller thread or in extra threadpool, compare throughputs/latencies).

Btw, the current code will default the ThreadPoolExecutor's max_workers to the number of processors, which isn't what we want (https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor). In contrast, the openapi client generated code defaults it to some constant multiplied by the number of processors (5\*numproc? 10\*numproc? I forget). If we end up going with the extra threadpool, we should fix this.

rajat08:
> @rajat08 how does the chaning behave in case of errors? Can you test? Does it affect the error messages presented to customers?

I actually need to change something for error messages, forgot about it for a second. Thanks

pinecone/core/grpc/index_grpc.py
rajat08: I put in a thread pool on the assumption it would hurt performance otherwise, though I didn't go far enough to check actual numbers.

Also yeah, if we keep the threadpool, changing default workers would be good.

PR #51, opened 2022-04-18
Add delete by metadata filter support

benjaminran:
Adds support for deleting by metadata filter


benjaminran:
Still need to add changelog entry

specs/pinecone_api.json
```diff
@@ -121,7 +121,7 @@
       "delete": {
         "summary": "Delete",
         "description": "The `Delete` operation deletes vectors, by id, from a single namespace.\nYou can delete items by their id, from a single namespace.",
-        "operationId": "delete",
+        "operationId": "delete1",
```
rajat08, line 124: Did you mean this to be delete1 or leftover from testing?

rajat08:
@benjaminran we would need to update index_grpc too

specs/pinecone_api.json
benjaminran: Intentional - that's automatically generated (it's the old DELETE api, now superceded by the new POST api for /vectors/delete). We could override the operationId if we want, but I thought delete1 was reasonable as is.

benjaminran:
Thanks @rajat08 , I've updated grpc_index

pinecone/core/grpc/index_grpc.py
```diff
@@ -259,7 +259,11 @@ def _vector_transform(item):
             return self._wrap_grpc_call(self.stub.Upsert, request, timeout=timeout)
 
     def delete(self, *args, async_req=False, **kwargs):
-        request = DeleteRequest(*args, **kwargs)
+        _filter = dict_to_proto_struct(kwargs.pop('filter', None))
+        filter_param = {}
+        if _filter:
+            filter_param['filter'] = _filter
```
rajat08, line 265: @benjaminran  is the param expected to be something like {'filter':'...'} ?

benjaminran: It's expected to be a filter= kwarg, like `index.delete(filter={"genre":"documentary"})`. 

I'm using filter_param here to avoid passing the filter= kwarg as struct({}) if the user didn't specify it (so that struct({}) isn't added to the payload). This is because
- it's not adding any value and just adds a bit of size to the payload
- and the backend enforces that filter shouldn't be passed if ids or delete_all==true is; I'd have to add extra logic in the backend to consider struct({}) a value that doesn't count as a filter

PR #52, opened 2022-04-27
Sc 13480/deprecate batch query api

yarden-slon:


mutayroei:
@yarden-slon I wonder if we want to completely remove the unary query from the client, such that new versions won't have the option to send batch queries. WDYT?

yarden-slon:
> @yarden-slon I wonder if we want to completely remove the unary query from the client, such that new versions won't have the option to send batch queries. WDYT?

I think we need to keep it for now and remove it after some adaptation period

pinecone/core/client/model/query_request.py
```diff
@@ -166,9 +169,11 @@ def _from_openapi_data(cls, top_k, queries, *args, **kwargs):  # noqa: E501
...
                                 _visited_composed_classes = (Animal,)
             namespace (str): The namespace to query.. [optional]  # noqa: E501
-            filter ({str: (bool, date, datetime, dict, float, int, list, str, none_type)}): The filter to apply if no filter is specified in `QueryVector. You can use vector metadata to limit your search. See https://www.pinecone.io/docs/metadata-filtering/.. [optional]  # noqa: E501
+            filter ({str: (bool, date, datetime, dict, float, int, list, str, none_type)}): The filter to apply. You can use vector metadata to limit your search. See https://www.pinecone.io/docs/metadata-filtering/.. [optional]  # noqa: E501
             include_values (bool): Indicates whether vector values are included in the response.. [optional] if omitted the server will use the default value of False  # noqa: E501
             include_metadata (bool): Indicates whether metadata is included in the response as well as the ids.. [optional] if omitted the server will use the default value of False  # noqa: E501
+            queries ([QueryVector]): The query vectors.. [optional]  # noqa: E501
```
mutayroei, line 175: it's strange to have them bother here, with the same documentation. At least let's write that it's deprecated?

pinecone/core/client/model/query_request.py
```diff
@@ -257,9 +260,11 @@ def __init__(self, top_k, queries, *args, **kwargs):  # noqa: E501
...
```
mutayroei, line 266: same here

mutayroei:
Can you attach an example of how a query response looks like after this change?

yarden-slon:
> Can you attach an example of how a query response looks like after this change?
For unary query:
{'matches': [{'id': 'vec', 'score': 1.0024, 'values': []}],
 'namespace': 'test_query_simple',
 'results': []}
 For batch query: 
{'matches': [],
 'namespace': '',
 'results': [{'matches': [{'id': 'vec', 'score': 1.0024, 'values': []}],
 'namespace': 'test_query_simple',}]}

PR #53, opened 2022-04-27
Sanitize numpy arrays

filipecaixeta:
Including numpy arrays in sanitize_for_serialization


pinecone/core/client/api_client.py
```diff
@@ -43,6 +43,12 @@
     validate_and_convert_types
 )
 
+try:
```
filipecaixeta, line 46: I don't want to include numpy as a dependency only because I need to do a type check

filipecaixeta:
@rajat08 can you look at this?

rajat08:
@filipecaixeta  Thanks for bringing this up! 

I think the NumPy dependency and the associated functions are leftovers from when we used NumPy arrays instead of Lists for vectors. I think the better thing to do here would be to remove the functions and dependencies altogether. 

Since api_client.py is created by the OpenAPI generator, we try not to make manual changes to this file.

I'll check in with the team once and get rid of NumPy if possible in the next release. Thanks again!

filipecaixeta:
I didn't notice you already have numpy as a dependency.
My point here is that I can't index or query numpy arrays without converting them to a python list so the PR was more about that than the numpy dependency

rajat08:
> 

Got it, thanks!
Even in that case, I don't think it's a good idea to manually edit files that are auto-generated. Our backend does not accept NumPy arrays right now. I would suggest making changes in index.py and index_grpc.py and adding helper functions in places like [this](https://github.com/pinecone-io/pinecone-python-client/blob/60319f2dc3b854f8a6e2acd3d1b7223a60551d0a/pinecone/core/grpc/index_grpc.py#L244) to do the conversion explicitly.

rajat08:
Closing this PR since we do not want to manually edit the autogenerated code to support this. 
Thanks, @filipecaixeta for bringing this to our attention!

PR #92, opened 2022-08-08
Update Code

pineconemachine:
Update generated code from latest proto defintions


pinecone/manage.py
```diff
@@ -199,3 +212,40 @@ def scale_index(name: str, replicas: int):
...
+
+
+def list_collections():
+    """List all collections"""
+    api_instance = _get_api_instance()
+    response = api_instance.list_collections()
+    return response
+
+
+def delete_collction(name: str):
```
jackpertschuk, line 236: typo?

jackpertschuk: @rajat08

rajat08: Yep, let me fix it

PR #98, opened 2022-08-14
Remove index_config from IndexDescription

yarden-slon:


pinecone/manage.py
```diff
@@ -201,7 +201,7 @@ def describe_index(name: str):
     return IndexDescription(name=db['name'], metric=db['metric'],
                             replicas=db['replicas'], dimension=db['dimension'], shards=db['shards'],
                             pods=db.get('pods', db['shards'] * db['replicas']), pod_type=db.get('pod_type', 'p1'),
-                            index_config=db['index_config'], status={'ready': ready, 'state': state},
+                            index_config=db.get('index_config'), status={'ready': ready, 'state': state},
```
rajat08, line 204: This may still show up as None in the response. Should we just get rid of it?

rajat08 (approved):

pinecone/manage.py
mutayroei: good point, let's just remove

PR #99, opened 2022-08-15
Change how IndexDescription works

rajat08:


rajat08:
@mutayroei  @yarden-slon  wdyt

mutayroei:
Hmm it adds another level of 'database' to the response, doesn't it?

rajat08:
> Hmm it adds another level of 'database' to the response, doesn't it?

good catch, that can be removed, the status needs to be sanitized too. Apart from that the approach may work

pinecone/manage.py
```diff
@@ -194,14 +190,14 @@ def describe_index(name: str):
     """
     api_instance = _get_api_instance()
     response = api_instance.describe_index(name)
-    db = response['database']
-    ready = response['status']['ready']
-    state = response['status']['state']
-    return IndexDescription(name=db['name'], metric=db['metric'],
-                            replicas=db['replicas'], dimension=db['dimension'], shards=db['shards'],
-                            pods=db.get('pods', db['shards'] * db['replicas']), pod_type=db.get('pod_type', 'p1'),
```
yarden-slon, line 202: Do we want to expose `shards` and `replicas` if they are in the dict?  or should we handle it like it used to?

pinecone/manage.py
```diff
@@ -194,14 +190,14 @@ def describe_index(name: str):
...
-                            status={'ready': ready, 'state': state}, metadata_config=db.get('metadata_config'),
-                            source_collection=db.get('source_collection', ''))
+    db_status = response.get('status', {})
+    response = response['database'].to_dict()
+    state = db_status.get('state', '')
+    ready = db_status.get('ready', '')
+    status = {'state': state, 'ready': ready}
+    response['status'] = status
+    response_object = IndexDescription(response.keys(), response.values())
```
yarden-slon, line 199: Is it better to just get the dict as a parameter to `IndexDescription` __init__?

pinecone/manage.py
rajat08: good point, forgot about that. I will replace it with pods

pinecone/manage.py
rajat08: yeah can just send the dict too here. anything should be okay

pinecone/manage.py
```diff
@@ -194,14 +190,14 @@ def describe_index(name: str):
...
-    ready = response['status']['ready']
-    state = response['status']['state']
-    return IndexDescription(name=db['name'], metric=db['metric'],
-                            replicas=db['replicas'], dimension=db['dimension'], shards=db['shards'],
-                            pods=db.get('pods', db['shards'] * db['replicas']), pod_type=db.get('pod_type', 'p1'),
-                            status={'ready': ready, 'state': state}, metadata_config=db.get('metadata_config'),
-                            source_collection=db.get('source_collection', ''))
+    db_status = response.get('status', {})
+    response = response['database'].to_dict()
+    state = db_status.get('state', '')
```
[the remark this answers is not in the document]
rajat08: Hmmm okay, i was thinking about what would be a good thing to do when the fields are deprecated in the future. Maybe we should stick to existing data types rather than None. just thinking out loud

pinecone/manage.py
rajat08: Also, we have been exposing shards and replicas always, just added pods for backward compatibility. We can continue doing that

yarden-slon: Oh right, missed that

rajat08:
@yarden-slon  this change would need changes to test_get on the ControlPlane, the response is an object instead of NamedTuple, I can make those chagnes

yarden-slon:
> @yarden-slon this change would need changes to test_get on the ControlPlane, the response is an object instead of NamedTuple, I can make those chagnes

Yes great thank you

yarden-slon (approved):

benjaminran:
Looks like it doesn't give a very useful default __repr__() - could you fix this?
```
>>> pinecone.describe_collection('colbench-col-r071-d512-nv1mm-p2-euc-sh1')
<pinecone.manage.CollectionDescription object at 0x1094de880>
>>> pinecone.describe_collection('colbench-col-r071-d512-nv1mm-p2-euc-sh1').__dict__
{'name': 'colbench-col-r071-d512-nv1mm-p2-euc-sh1', 'status': 'Initializing', 'dimension': 512.0}
```

PR #101, opened 2022-08-16
Make dimension optional on index creation

pineconemachine:
Update generated code from latest proto defintions

### DIFF
--- a/pinecone/core/client/model/index_meta_database.py
+++ b/pinecone/core/client/model/index_meta_database.py
@@ -93,7 +93,7 @@ class IndexMetaDatabase(ModelNormal):
         lazy_import()
         return {
             'name': (str,),  # noqa: E501
-            'dimensions': (str,),  # noqa: E501
+            'dimension': (str,),  # noqa: E501
             'index_type': (str,),  # noqa: E501
             'metric': (str,),  # noqa: E501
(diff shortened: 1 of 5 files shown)


pinecone/core/client/model/index_meta_database.py
```diff
@@ -164,7 +164,7 @@ def _from_openapi_data(cls, *args, **kwargs):  # noqa: E501
                                 through its discriminator because we passed in
                                 _visited_composed_classes = (Animal,)
             name (str): [optional]  # noqa: E501
-            dimensions (str): [optional]  # noqa: E501
+            dimension (str): [optional]  # noqa: E501
```
mutayroei, line 167: why is it (str)?

pinecone/core/client/model/index_meta_database.py
```diff
@@ -255,7 +255,7 @@ def __init__(self, *args, **kwargs):  # noqa: E501
...
```
mutayroei, line 258: same here

pinecone/core/client/model/index_meta_database.py
rajat08: yeah needs to change in the spec, we don't use it so maybe skip it for this release?

PR #109, opened 2022-11-02
Add sparse values support for vectors

mutayroei:
While this PR technically does contain the auto-generated python code that calls the Text (BM25) API - this code **isn't** wired into the `Index` object so it's not exposed to users.


pinecone/core/client/api/hybrid_operations_api.py
```diff
@@ -0,0 +1,407 @@
+#
```
igiloh-pinecone, line 1: @acatav @miararoy I'm not sure we should push this file to the client repo, which is exposed to customers.  
Since it's auto-generated code - we can always commit it later when we also wire it to the `Index` object.   

For now these files would just be sitting there (and installed to customers' machines who `pip install pinecone`) for no reason.

tests/unit/test_index.py
```diff
@@ -30,8 +32,38 @@ def test_upsert_request_tuples_id_data_metadata(mocker):
...
+    index.upsert([('vec1', vals1, md1, sparse1),
+                  ('vec2', vals2, md2, sparse2)])
+    index._vector_api.upsert.assert_called_once_with(
+        pinecone.UpsertRequest(vectors=[
+            pinecone.Vector(id='vec1', values=vals1, metadata=md1, sparse_values=sparse1),
+            pinecone.Vector(id='vec2', values=vals2, metadata=md2, sparse_values=sparse2)
+        ])
+    )
+
+
```
igiloh-pinecone, line 55: I would add some negative tests as well - for example a test where you pass a `Dict[int, float]` to the 3rd param in the tuple, and `assert` that a proper exception is raised

igiloh-pinecone (approved):
@acatav I only had one suggestion regarding a few additional test cases. Other than that LGTM.

I'm approving anyway. See if you'd like to add these tests or not, and feel free to merge.

PR #114, opened 2022-12-22
Fix "Connection Reset by peer" error after long idle periods

igiloh-pinecone:
It seems that OpenAPI's inherent Rest client uses connection pooling without a keep alive mechanism.  
In some cases, the client keeps holding an idle connection for long periods (>15 min), yet the server (Pinecone's load balancer) seem to be terminating the connection after ~10 minutes of idleness.  

I've configured OpenAPI's client to enable Keep Alive for all connections. Since the LB is configured to terminate the connection regardless of keep alive, this won't lead to having many dangling connections.  
The main benefit is having the client "finding out" early about the dropped connection, so it won't try to use it for the next user request.  

I've tested the solution (manually) on Both Mac OS, Linux and Windows Server OSes - and it seems to resolve the failure.

### DIFF
--- a/pinecone/config.py
+++ b/pinecone/config.py
@@ -186,8 +186,11 @@ class _CONFIG:
             socket_params += [(socket.IPPROTO_TCP, socket.TCP_KEEPCNT, keep_alive_tries)]
 
         # TCP Keep Alive Probes for Windows OS
-        elif platform == 'win32' and hasattr(socket, "SIO_KEEPALIVE_VALS"):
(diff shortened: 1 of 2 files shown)


igiloh-pinecone:
@rajat08 @benjaminran FYI. Feel free to review \ comment

rajat08:
this looks great @igiloh-pinecone , do we want to make these settings configurable? I think that will be helpful

acatav (approved):

benjaminran (approved):
🙌

PR #116, opened 2022-12-22
Refactor API args and add documentation 

acatav:
This PR cleans current code while keeping 100% backwards compatibility, by:
1. add explicit arguments and types
2. add documentation
3. add unit tests for all arguments combinations

### DIFF
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -311,7 +311,7 @@ class GRPCIndex(GRPCIndexBase):
                ids: Optional[List[str]] = None,
                delete_all: Optional[bool] = None,
                namespace: Optional[str] = None,
-               filter: Optional[Dict[str, Union[str, float, int, bool, List, Dict]]] = None,
+               filter: Optional[Dict[str, Union[str, float, int, bool, List, dict]]] = None,
                async_req: bool = False,
                **kwargs) -> Union[DeleteResponse, PineconeGrpcFuture]:
         """
@@ -336,7 +336,7 @@ class GRPCIndex(GRPCIndexBase):
                                Default is False.
             namespace (str): The namespace to delete vectors from [optional]
                              If not specified, the default namespace is used.
-            filter (Dict[str, Union[str, float, int, bool, List, Dict]]):
+            filter (Dict[str, Union[str, float, int, bool, List, dict]]):
                     If specified, the metadata filter here will be used to select the vectors to delete.
                     This is mutually exclusive with specifying ids to delete in the ids param or using delete_all=True.
                      See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
@@ -396,7 +396,7 @@ class GRPCIndex(GRPCIndexBase):
               queries: Optional[Union[List[GRPCQueryVector], List[Tuple]]] = None,
               namespace: Optional[str] = None,
               top_k: Optional[int] = None,
-              filter: Optional[Dict[str, Union[str, float, int, bool, List, Dict]]] = None,
+              filter: Optional[Dict[str, Union[str, float, int, bool, List, dict]]] = None,
               include_values: Optional[bool] = None,
               include_metadata: Optional[bool] = None,
               **kwargs) -> QueryResponse:
@@ -423,7 +423,7 @@ class GRPCIndex(GRPCIndexBase):
             top_k (int): The number of results to return for each query. Must be an integer greater than 1.
             namespace (str): The namespace to fetch vectors from.
                              If not specified, the default namespace is used. [optional]
-            filter (Dict[str, Union[str, float, int, bool, List, Dict]]):
+            filter (Dict[str, Union[str, float, int, bool, List, dict]]):
                     The filter to apply. You can use vector metadata to limit your search.
                     See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
             include_values (bool): Indicates whether vector values are included in the response.
@@ -511,7 +511,7 @@ class GRPCIndex(GRPCIndexBase):
             return self._wrap_grpc_call(self.stub.Update, request, timeout=timeout)
 
     def describe_index_stats(self,
-                             filter: Optional[Dict[str, Union[str, float, int, bool, List, Dict]]] = None,
+                             filter: Optional[Dict[str, Union[str, float, int, bool, List, dict]]] = None,
                              **kwargs) -> DescribeIndexStatsResponse:
         """
         The DescribeIndexStats operation returns statistics about the index's contents.
@@ -522,7 +522,7 @@ class GRPCIndex(GRPCIndexBase):
             >>> index.describe_index_stats(filter={'key': 'value'})
 
         Args:
-            filter (Dict[str, Union[str, float, int, bool, List, Dict]]):
+            filter (Dict[str, Union[str, float, int, bool, List, dict]]):
             If this parameter is present, the operation only returns statistics for vectors that satisfy the filter.
             See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
 
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -281,11 +281,10 @@ class GRPCIndex(GRPCIndexBase):
 
                     Note: the dimension of each vector must match the dimension of the index.
             async_req (bool): If True, the upsert operation will be performed asynchronously.
-                              Defaults to False. [optional]
+                              Defaults to False. See: https://docs.pinecone.io/docs/performance-tuning [optional]
             namespace (str): The namespace to write to. If not specified, the default namespace is used. [optional]
 
         Returns: UpsertResponse, contains the number of vectors upserted
-                 or a PineconeGrpcFuture object if async_req is True.
         """
 
         args_dict = self._parse_args_to_dict([('namespace', namespace)])
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -274,7 +274,8 @@ class GRPCIndex(GRPCIndexBase):
                         Examples: ('id1', [1.0, 2.0, 3.0], {'key': 'value'}), ('id2', [1.0, 2.0, 3.0])
 
                     2) if a GRPCVector object is used, a GRPCVector object must be of the form
-                        GRPCVector(id, values, metadata), where metadata is an optional argument.
+                        GRPCVector(id, values, metadata), where metadata is an optional argument of type
+                        Dict[str, Union[str, float, int, bool, List[int], List[float], List[str]]]
                        Examples: GRPCVector(id='id1', values=[1.0, 2.0, 3.0], metadata={'key': 'value'}),
                                  GRPCVector(id='id2', values=[1.0, 2.0, 3.0])
 
@@ -310,7 +311,7 @@ class GRPCIndex(GRPCIndexBase):
                ids: Optional[List[str]] = None,
                delete_all: Optional[bool] = None,
                namespace: Optional[str] = None,
-               filter: Optional[Dict[str, Any]] = None,
+               filter: Optional[Dict[str, Union[str, float, int, bool, List, Dict]]] = None,
                async_req: bool = False,
                **kwargs) -> Union[DeleteResponse, PineconeGrpcFuture]:
         """
@@ -335,7 +336,7 @@ class GRPCIndex(GRPCIndexBase):
                                Default is False.
             namespace (str): The namespace to delete vectors from [optional]
                              If not specified, the default namespace is used.
-            filter (Dict[str, Union[bool, date, dict, float, int, list, str, none_type)]]):
+            filter (Dict[str, Union[str, float, int, bool, List, Dict]]):
                     If specified, the metadata filter here will be used to select the vectors to delete.
                     This is mutually exclusive with specifying ids to delete in the ids param or using delete_all=True.
                      See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
@@ -395,7 +396,7 @@ class GRPCIndex(GRPCIndexBase):
               queries: Optional[Union[List[GRPCQueryVector], List[Tuple]]] = None,
               namespace: Optional[str] = None,
               top_k: Optional[int] = None,
-              filter: Optional[Dict[str, Any]] = None,
+              filter: Optional[Dict[str, Union[str, float, int, bool, List, Dict]]] = None,
               include_values: Optional[bool] = None,
               include_metadata: Optional[bool] = None,
               **kwargs) -> QueryResponse:
@@ -422,7 +423,7 @@ class GRPCIndex(GRPCIndexBase):
             top_k (int): The number of results to return for each query. Must be an integer greater than 1.
             namespace (str): The namespace to fetch vectors from.
                              If not specified, the default namespace is used. [optional]
-            filter (Dict[str, Union[bool, date, dict, float, int, list, str, none_type)]]):
+            filter (Dict[str, Union[str, float, int, bool, List, Dict]]):
                     The filter to apply. You can use vector metadata to limit your search.
                     See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
             include_values (bool): Indicates whether vector values are included in the response.
@@ -469,7 +470,8 @@ class GRPCIndex(GRPCIndexBase):
                id: str,
                async_req: bool = False,
                values: Optional[List[float]] = None,
-               set_metadata: Optional[Dict[str, Any]] = None,
+               set_metadata: Optional[Dict[str,
+                                           Union[str, float, int, bool, List[int], List[float], List[str]]]] = None,
                namespace: Optional[str] = None,
                **kwargs) -> Union[UpdateResponse, PineconeGrpcFuture]:
         """
@@ -487,7 +489,7 @@ class GRPCIndex(GRPCIndexBase):
             async_req (bool): If True, the update operation will be performed asynchronously.
                               Defaults to False. [optional]
             values (List[float]): vector values to set. [optional]
-            set_metadata (Dict[str, Union[bool, date, dict, float, int, list, str, none_type)]]):
+            set_metadata (Dict[str, Union[str, float, int, bool, List[int], List[float], List[str]]]]):
                 metadata to set for vector. [optional]
             namespace (str): Namespace name where to update the vector.. [optional]
 
@@ -509,7 +511,7 @@ class GRPCIndex(GRPCIndexBase):
             return self._wrap_grpc_call(self.stub.Update, request, timeout=timeout)
 
     def describe_index_stats(self,
-                             filter: Dict[str, Any] = {},
+                             filter: Optional[Dict[str, Union[str, float, int, bool, List, Dict]]] = None,
                              **kwargs) -> DescribeIndexStatsResponse:
         """
         The DescribeIndexStats operation returns statistics about the index's contents.
@@ -520,16 +522,18 @@ class GRPCIndex(GRPCIndexBase):
             >>> index.describe_index_stats(filter={'key': 'value'})
 
         Args:
-            filter (Dict[str, Union[bool, date, dict, float, int, list, str, none_type)]]):
+            filter (Dict[str, Union[str, float, int, bool, List, Dict]]):
             If this parameter is present, the operation only returns statistics for vectors that satisfy the filter.
             See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
 
         Returns: DescribeIndexStatsResponse object which contains stats about the index.
         """
-
-        filter = dict_to_proto_struct(filter)
+        if filter is not None:
+            filter = dict_to_proto_struct(filter)
+        args_dict = self._parse_args_to_dict([('filter', filter)])
         timeout = kwargs.pop('timeout', None)
-        request = DescribeIndexStatsRequest(filter=filter)
+
+        request = DescribeIndexStatsRequest(**args_dict)
         response = self._wrap_grpc_call(self.stub.DescribeIndexStats, request, timeout=timeout)
         json_response = json_format.MessageToDict(response)
         return parse_stats_response(json_response)
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -297,8 +297,9 @@ class GRPCIndex(GRPCIndexBase):
                 return GRPCVector(id=id, values=values, metadata=dict_to_proto_struct(metadata) or {})
             raise ValueError(f"Invalid vector value passed: cannot interpret type {type(item)}")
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -72,8 +72,9 @@ import pinecone
 
 pinecone.init(api_key="YOUR_API_KEY", environment="us-west1-gcp")
 
-index_description = pinecone.describe_index("example-index"
-```)
+index_description = pinecone.describe_index("example-index")
+
```
miararoy, line 76: should consider heavier refactoring of the Readme (not at this PR, in general)

pinecone/index.py
```diff
@@ -56,8 +62,47 @@ def __init__(self, index_name: str, pool_threads=1):
         self._vector_api = VectorOperationsApi(self)
 
     @validate_and_convert_errors
-    def upsert(self, vectors, **kwargs):
+    def upsert(self,
+               vectors: Union[List[Vector], List[Tuple]],
+               namespace: Optional[str] = None,
```
miararoy, line 67: This is actually UpsertRequest if you want to be religious about it :)

pinecone/index.py
```diff
@@ -70,31 +115,169 @@ def _vector_transform(item):
...
+        """
+        args_dict = self._parse_args_to_dict([('namespace', namespace)])
+        return self._vector_api.fetch(ids=ids, **args_dict, **kwargs)
 
     @validate_and_convert_errors
-    def query(self, vector=[], id='', queries=[], **kwargs):
+    def query(self,
+              vector: Optional[List[float]] = None,
+              id: Optional[str] = None,
+              queries: Optional[Union[List[QueryVector], List[Tuple]]] = None,
```
miararoy, line 214: this is deprecated, are we keeping it only for backward compatibility?

miararoy (approved):

pinecone/index.py
```diff
@@ -70,31 +115,169 @@ def _vector_transform(item):
...
+        The returned vectors include the vector data and/or metadata.
+
+        API reference: https://docs.pinecone.io/reference/fetch
+
+        Args:
+            ids (List[str]): The vector IDs to fetch.
+            namespace (str): The namespace to fetch vectors from.
+                             If not specified, the default namespace is used. [optional]
+        Keyword Args:
+        Supports OpenAPI client keyword arguments. See pinecone.core.client.models.FetchResponse for more details.
```
gdj0nes, line 198: Indent 
    Supports...

pinecone/index.py
```diff
@@ -56,8 +62,47 @@ def __init__(self, index_name: str, pool_threads=1):
...
+                                 Vector(id='id2', values=[1.0, 2.0, 3.0])
+
+                    Note: the dimension of each vector must match the dimension of the index.
+
+            namespace (str): The namespace to write to. If not specified, the default namespace is used. [optional]
+
+        Keyword Args:
+            Supports OpenAPI client keyword arguments. See pinecone.core.client.models.UpsertRequest for more details.
+
+        Returns: UpsertResponse object which basically contains the number of vectors upserted.
```
gdj0nes, line 95: UpsertResponse, includes the number of vectors upserted.

pinecone/core/grpc/index_grpc.py
```diff
@@ -251,11 +245,50 @@ def traceback(self, timeout=None):
...
+                        GRPCVector(id, values, metadata), where metadata is an optional argument.
+                       Examples: GRPCVector(id='id1', values=[1.0, 2.0, 3.0], metadata={'key': 'value'}),
+                                 GRPCVector(id='id2', values=[1.0, 2.0, 3.0])
+
+                    Note: the dimension of each vector must match the dimension of the index.
+            async_req (bool): If True, the upsert operation will be performed asynchronously.
+                              Defaults to False. [optional]
+            namespace (str): The namespace to write to. If not specified, the default namespace is used. [optional]
+
+        Returns: UpsertResponse object which basically contains the number of vectors upserted
```
gdj0nes, line 281: UpsertResponse, contains the number of vectors upserted

pinecone/core/grpc/index_grpc.py
```diff
@@ -251,11 +245,50 @@ def traceback(self, timeout=None):
...
+                 or a PineconeGrpcFuture object if async_req is True.
+
+        Call Examples:
+            index.upsert([('id1', [1.0, 2.0, 3.0], {'key': 'value'}), ('id2', [1.0, 2.0, 3.0])], namespace='ns1')
+            index.upsert([GRPCVector(id='id1', values=[1.0, 2.0, 3.0], metadata={'key': 'value'}),
+                          GRPCVector(id='id2', values=[1.0, 2.0, 3.0])], async_req=True)
```
gdj0nes, line 287: update with >>>

pinecone/core/grpc/index_grpc.py
```diff
@@ -264,37 +297,141 @@ def _vector_transform(item):
...
+                     See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
+            async_req (bool): If True, the delete operation will be performed asynchronously.
+                              Defaults to False. [optional]
+
+        Returns: DeleteResponse (contains no data) or a PineconeGrpcFuture object if async_req is True.
+
+        Call Examples:
+            index.delete(ids=['id1', 'id2'], namespace='my_namespace')
+            index.delete(delete_all=True, namespace='my_namespace')
+            index.delete(filter={'key': 'value'}, namespace='my_namespace', async_req=True)
```
gdj0nes, line 344: Update with >>>

pinecone/core/grpc/index_grpc.py
```diff
@@ -264,37 +297,141 @@ def _vector_transform(item):
...
+        Args:
+            ids (List[str]): The vector IDs to fetch.
+            namespace (str): The namespace to fetch vectors from.
+                             If not specified, the default namespace is used. [optional]
+
+        Returns: FetchResponse object which contains the list of Vector objects, and namespace name.
+
+        Call Examples:
+            index.fetch(ids=['id1', 'id2'], namespace='my_namespace')
+            index.fetch(ids=['id1', 'id2'])
```
gdj0nes, line 380: Update with >>>

pinecone/core/grpc/index_grpc.py
```diff
@@ -264,37 +297,141 @@ def _vector_transform(item):
...
 
-    def query(self, vector=[], id='', queries=[], **kwargs):
-        timeout = kwargs.pop('timeout', None)
-
+    def query(self,
+              vector: Optional[List[float]] = None,
+              id: Optional[str] = None,
+              queries: Optional[Union[List[GRPCQueryVector], List[Tuple]]] = None,
+              namespace: Optional[str] = None,
+              top_k: Optional[int] = None,
```
gdj0nes, line 396: top_k is not optional

pinecone/core/grpc/index_grpc.py
```diff
@@ -264,37 +297,141 @@ def _vector_transform(item):
...
+                                     If omitted the server will use the default value of False  [optional]
+
+        Returns: QueryResponse object which contains the list of the closest vectors as ScoredVector objects,
+                 and namespace name.
+
+        Call Examples:
+            index.query(vector=[1, 2, 3], top_k=10, namespace='my_namespace')
+            index.query(id='id1', top_k=10, namespace='my_namespace')
+            index.query(vector=[1, 2, 3], top_k=10, namespace='my_namespace', filter={'key': 'value'})
+            index.query(id='id1', top_k=10, namespace='my_namespace', include_metadata=True, include_values=True)
```
gdj0nes, line 433: Add >>>

pinecone/core/grpc/index_grpc.py
```diff
@@ -306,36 +443,96 @@ def _query_transform(item):
...
+        Args:
+            filter (Dict[str, Union[bool, date, dict, float, int, list, str, none_type)]]):
+            If this parameter is present, the operation only returns statistics for vectors that satisfy the filter.
+            See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
+
+        Returns: DescribeIndexStatsResponse object which contains stats about the index.
+
+        Call Examples:
+            index.describe_index_stats()
+            index.describe_index_stats(filter={'key': 'value'})
```
gdj0nes, line 526: Add >>>

pinecone/index.py
```diff
@@ -38,6 +39,11 @@ def parse_query_response(response: QueryResponse, unary_query: bool):
 
 class Index(ApiClient):
```
gdj0nes, line 41: Remove space between doc string and index

pinecone/index.py
```diff
@@ -38,6 +39,11 @@ def parse_query_response(response: QueryResponse, unary_query: bool):
...
 
+    """
+    A client for interacting with a Pinecone index via REST API.
+    For improved performance, use the Pinecone GRPC index client (https://docs.pinecone.io/docs/performance-tuning)
```
gdj0nes, line 44: I'm not sure we should link this here

pinecone/index.py
```diff
@@ -70,31 +115,169 @@ def _vector_transform(item):
...
+              vector: Optional[List[float]] = None,
+              id: Optional[str] = None,
+              queries: Optional[Union[List[QueryVector], List[Tuple]]] = None,
+              top_k: Optional[int] = None,
+              namespace: Optional[str] = None,
+              filter: Optional[Dict[str, Any]] = None,
+              include_values: Optional[bool] = None,
+              include_metadata: Optional[bool] = None,
+              **kwargs) -> QueryResponse:
+
```
gdj0nes, line 221: Remove space

gdj0nes (changes requested):

gdj0nes:
We should move examples higher in the docs strings right under the description and use >>> for the code sections. Are we following a particular style guide?

```
    def upsert(self,
               vectors: Union[List[Vector], List[Tuple]],
               namespace: Optional[str] = None,
               **kwargs) -> UpsertResponse:
        """The upsert operation writes vectors into a namespace.
        If a new value is upserted for an existing vector id, it will overwrite the previous value.
        API reference: https://docs.pinecone.io/reference/upsert

        Examples:
            >>> index.upsert([('id1', [1.0, 2.0, 3.0], {'key': 'value'}), ('id2', [1.0, 2.0, 3.0])])
            >>> index.upsert([Vector(id='id1', values=[1.0, 2.0, 3.0], metadata={'key': 'value'}), Vector(id='id2', values=[1.0, 2.0, 3.0])])

        Args:
            vectors (Union[List[Vector], List[Tuple]]): A list of vectors to upsert.
                     A vector can be represented by a 1) Vector object or a 2) tuple.
                     1) if a tuple is used, it must be of the form (id, values, metadata) or (id, values).
                        where id is a string, vector is a list of floats, and metadata is a dict.
                        Examples: ('id1', [1.0, 2.0, 3.0], {'key': 'value'}), ('id2', [1.0, 2.0, 3.0])
                    2) if a Vector object is used, a Vector object must be of the form Vector(id, values, metadata),
                        where metadata is an optional argument.
                       Examples: Vector(id='id1', values=[1.0, 2.0, 3.0], metadata={'key': 'value'}),
                                 Vector(id='id2', values=[1.0, 2.0, 3.0])
                    Note: the dimension of each vector must match the dimension of the index.
            namespace (str): The namespace to write to. If not specified, the default namespace is used. [optional]

        Keyword Args:
            Supports OpenAPI client keyword arguments. See pinecone.core.client.models.UpsertRequest for more details.

        Returns: UpsertResponse object which basically contains the number of vectors upserted.
        """
```

pinecone/index.py
acatav: Actually I think it's indeed a list of vectors, the UpsertRequest object is created internally, right?

pinecone/index.py
acatav: yes, we don't want to do anything in this version that might break code for someone

pinecone/core/grpc/index_grpc.py
acatav: You are right, but since until now it was in the kwargs, I didn't have a good place to put it as arg without breaking backwards compatibility (and we decided that this PR should preserve backwards compatibility). Do you have an idea how to make it a required arg?

pinecone/index.py
```diff
@@ -70,31 +115,169 @@ def _vector_transform(item):
...
+        args_dict = self._parse_args_to_dict([('namespace', namespace)])
+        return self._vector_api.fetch(ids=ids, **args_dict, **kwargs)
 
     @validate_and_convert_errors
-    def query(self, vector=[], id='', queries=[], **kwargs):
+    def query(self,
+              vector: Optional[List[float]] = None,
+              id: Optional[str] = None,
+              queries: Optional[Union[List[QueryVector], List[Tuple]]] = None,
+              top_k: Optional[int] = None,
```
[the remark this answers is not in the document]
acatav: let's chat about that on the same comment from the grpc index

pinecone/index.py
```diff
@@ -108,34 +291,88 @@ def _query_transform(item):
...
+                             filter: Dict[str, Any] = {},
+                             **kwargs) -> DescribeIndexStatsResponse:
+        """
+        The DescribeIndexStats operation returns statistics about the index's contents.
+        For example: The vector count per namespace and the number of dimensions.
+
+        API reference: https://docs.pinecone.io/reference/describe_index_stats_post
+
+        Args:
+            filter (Dict[str, Union[bool, date, dict, float, int, list, str, none_type)]]):
```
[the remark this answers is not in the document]
acatav: For the docstring purposes, don't you think it will be more useful to put it like this even if it's not a proper type? BTW so what is the type of the date?

pinecone/index.py
```diff
@@ -108,34 +291,88 @@ def _query_transform(item):
...
+            If this parameter is present, the operation only returns statistics for vectors that satisfy the filter.
+            See https://www.pinecone.io/docs/metadata-filtering/.. [optional]
```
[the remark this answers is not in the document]
acatav: Not sure I understand this comment, can you please explain it further?

tox.ini
```diff
@@ -21,10 +21,11 @@ commands =
 
 [gh-actions]
 python =
-    3.6: py36
     3.7: py37
     3.8: py38
     3.9: py39
+    '3.10': py310
```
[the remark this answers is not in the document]
acatav: yes, if you don't put it like this it reads as python 3.1

.github/workflows/pr.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, '3.10', 3.11]
```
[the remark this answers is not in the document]
acatav: this in intentional, if you don't do it it reads as python 3.1

pinecone/index.py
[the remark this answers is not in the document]
gdj0nes: Supported types https://docs.pinecone.io/docs/metadata-filtering#supported-metadata-types

pinecone/index.py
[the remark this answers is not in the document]
gdj0nes: OK makes senese

gdj0nes (approved):

pinecone/core/grpc/index_grpc.py
```diff
@@ -251,11 +245,51 @@ def traceback(self, timeout=None):
...
 
+    """A client for interacting with a Pinecone index via GRPC API."""
+
     @property
     def stub_class(self):
         return VectorServiceStub
 
-    def upsert(self, vectors, async_req=False, **kwargs):
+    def upsert(self,
+               vectors: Union[List[GRPCVector], List[Tuple]],
```
igiloh-pinecone, line 255: You need to add `Vector` from `pinecone.core.model.vector` to the union

pinecone/core/grpc/index_grpc.py
```diff
@@ -264,37 +298,142 @@ def _vector_transform(item):
...
-        _filter = dict_to_proto_struct(kwargs.pop('filter', None))
-        filter_param = {}
-        if _filter:
-            filter_param['filter'] = _filter
-        request = DeleteRequest(*args, **kwargs, **filter_param)
+    def delete(self,
+               ids: Optional[List[str]] = None,
+               delete_all: Optional[bool] = None,
+               namespace: Optional[str] = None,
+               filter: Optional[Dict[str, Union[str, float, int, bool, List, Dict]]] = None,
```
igiloh-pinecone, line 314:
```suggestion
               filter: Optional[Dict[str, Union[str, float, int, bool, list, dict]]] = None,
```
You need to use the capital letter `Dict` only for composition (e.g. `Dict[str, str]`).  
For defining "any dict" in `typing` terms, the lower case `dict` is more correct

pinecone/core/grpc/index_grpc.py
```diff
@@ -251,11 +245,51 @@ def traceback(self, timeout=None):
...
+                       Examples: GRPCVector(id='id1', values=[1.0, 2.0, 3.0], metadata={'key': 'value'}),
+                                 GRPCVector(id='id2', values=[1.0, 2.0, 3.0])
+
+                    Note: the dimension of each vector must match the dimension of the index.
+            async_req (bool): If True, the upsert operation will be performed asynchronously.
+                              Defaults to False. [optional]
+            namespace (str): The namespace to write to. If not specified, the default namespace is used. [optional]
+
+        Returns: UpsertResponse, contains the number of vectors upserted
+                 or a PineconeGrpcFuture object if async_req is True.
```
igiloh-pinecone, line 288: Maybe we should add an async example to the `examples` section?  
So that users would know how to poll \ check the async response

igiloh-pinecone: Also - what happens if the user does `async_req=True` from the Rest API?  
Does he still get a `PineconeGrpcFuture` object? Isn't there some other object coming from OpenAPI?

pinecone/core/grpc/index_grpc.py
```diff
@@ -264,37 +298,142 @@ def _vector_transform(item):
                 return GRPCVector(id=id, values=values, metadata=dict_to_proto_struct(metadata) or {})
             raise ValueError(f"Invalid vector value passed: cannot interpret type {type(item)}")
 
-        request = UpsertRequest(vectors=list(map(_vector_transform, vectors)), **kwargs)
         timeout = kwargs.pop('timeout', None)
+
+        request = UpsertRequest(vectors=list(map(_vector_transform, vectors)), **args_dict, **kwargs)
```
igiloh-pinecone, line 303: I don't see why we need this `args_dict`.  
It's a different thing if we would have re-used it throughout the function many times. But we are only passing these params to a single function call.   

Why not pass the params explicitly?
```
request = UpsertRequest(vectors=list(map(_vector_transform, vectors)), namespace=namespace, **kwargs)
```
?

igiloh-pinecone (commented):
@acatav I haven't full reviewed all files, but I don't want to delay you too much.

Please see my suggestions so far.  
I will try to review the tests as well in the evening - but don't wait for me, feel free to merge.

tests/unit/test_grpc_index.py
```diff
@@ -0,0 +1,239 @@
+import pinecone
+from core.utils import dict_to_proto_struct
+from pinecone import DescribeIndexStatsRequest
+from pinecone.core.grpc.protos.vector_service_pb2 import Vector, DescribeIndexStatsRequest, UpdateRequest, \
+    UpsertRequest, FetchRequest, QueryRequest, DeleteRequest, QueryVector
+
+
+class TestRestIndex:
+
+    def setup_method(self):
```
igiloh-pinecone, line 10: Side note: we should probably use `UnitTest.TestCase` class.  
Let's defer this to a task in the future \ next versions.

tests/unit/test_grpc_index.py
```diff
@@ -0,0 +1,239 @@
+import pinecone
+from core.utils import dict_to_proto_struct
+from pinecone import DescribeIndexStatsRequest
+from pinecone.core.grpc.protos.vector_service_pb2 import Vector, DescribeIndexStatsRequest, UpdateRequest, \
+    UpsertRequest, FetchRequest, QueryRequest, DeleteRequest, QueryVector
+
+
+class TestRestIndex:
```
igiloh-pinecone, line 8: Change name to `GRPC`

tests/unit/test_grpc_index.py
```diff
@@ -0,0 +1,239 @@
...
+    def test_query_byVectorNoFilter_queryVectorNoFilter(self, mocker):
+        mocker.patch.object(self.index, '_wrap_grpc_call', autospec=True)
+        self.index.query(top_k=10, vector=self.vals1)
+        self.index._wrap_grpc_call.assert_called_once_with(
+            self.index.stub.Query,
+            QueryRequest(top_k=10, vector=self.vals1),
+            timeout=None,
+        )
+
+    def test_query_byVectorWithFilter_queryVectorWithFilter(self, mocker):
```
igiloh-pinecone, line 76: Consider checking also the return value, to make sure the `parse_query_response()` part works well

pinecone/core/grpc/index_grpc.py
acatav: I don't think it's possible to pass this vector, only the grpc one

pinecone/core/grpc/index_grpc.py
acatav: Didn't know that, cool

pinecone/core/grpc/index_grpc.py
acatav: The idea is to avoid passing default values if the user didn't provide something, so the server can have full control on the default params. I used the args_dict since the function creates it removes keys with None values from the dict

pinecone/core/grpc/index_grpc.py
igiloh-pinecone: Oh, wait, this is `GRPCIndex` 🤦 .  
You're right, I was thinking about the OpenAPI index. My bad

pinecone/core/grpc/index_grpc.py
igiloh-pinecone: I see. That's clever! 

Maybe we should change the function name to something like `parse_non_empty_args()` or `remove_none_args()` or something more explicit?

igiloh-pinecone: BTW - since these are optional params, I'm pretty sure that `OpenAPI` treats `None` as "empty" and fills in the appropriate default value anyway (on the server side).  
We can do a quick check to verify, and then we maybe don't even need this mechanism.

tests/unit/test_grpc_index.py
igiloh-pinecone: Deferred to [backlog](https://app.asana.com/0/1203260648987893/1203612700548377), resolving

pinecone/core/grpc/index_grpc.py
acatav: I linked to the documentation about that since it's rarely used

pinecone/core/grpc/index_grpc.py
acatav: I liked `parse_non_empty_args` better so renamed to that

PR #118, opened 2022-12-28
Use _check_type=False in GRPCIndex.query()

igiloh-pinecone:
We believe the type checking takes a long time which adds a lot of query latency.

I've added proper handling of the `_check_type` arg in `GRPCIndex.query()`, with default `False`.  
This argument is now correctly passed to `parse_query_response()`, which should save a lot of time for queries with large `top_k`

### DIFF
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -295,7 +295,6 @@ class GRPCIndex(GRPCIndexBase):
 
     def query(self, vector=[], id='', queries=[], **kwargs):
         timeout = kwargs.pop('timeout', None)
-        _check_type = kwargs.pop('_check_type', False)
 
         def _query_transform(item):
             if isinstance(item, GRPCQueryVector):
@@ -317,7 +316,7 @@ class GRPCIndex(GRPCIndexBase):
                                **{k: v for k, v in kwargs.items() if k in _QUERY_ARGS})
         response = self._wrap_grpc_call(self.stub.Query, request, timeout=timeout)
         json_response = json_format.MessageToDict(response)
-        return parse_query_response(json_response, vector or id, _check_type)
+        return parse_query_response(json_response, vector or id, _check_type=False)
 
     def update(self, id, async_req=False, **kwargs):
         _UPDATE_ARGS = ['values', 'set_metadata', 'namespace']
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -171,9 +171,10 @@ def parse_fetch_response(response: dict):
     return FetchResponse(vectors=vd, namespace=namespace, _check_type=False)
 
 
-def parse_query_response(response: dict, unary_query: bool):
+def parse_query_response(response: dict, unary_query: bool, _check_type: bool = False):
     res = []
 
+    # TODO: consider deleting this deprecated case
     for match in response.get('results', []):
         namespace = match.get('namespace', '')
         m = []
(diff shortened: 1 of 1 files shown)


harsha2010:
> We believe the type checking takes a long time which adds a lot of query latency.
> 
> I've added proper handling of the `_check_type` arg in `GRPCIndex.query()`, with default `False`. This argument is now correctly passed to `parse_query_response()`, which should save a lot of time for queries with large `top_k`

what is the evidence for this? is there a heat map that indicates the breakdown of latencies here?

igiloh-pinecone:
@harsha2010 great question!  
I was just about to post this analysis: 
image

On the left - the current code in `main`, in the middle - the suggested solution in this PR, on the right - temp workaround that completely bypasses parsing the response (hence returning to the user raw `pb2` object).  
This is for `top_k=50`. For larger top_ks we've seen even larger latency in the GPRC code path, all happening during this `_check_type` step (which is BTW an `OpenAPI` mechanism, we shouldn't even be using it on the GRPC path).

@miararoy FYI

pinecone/core/grpc/index_grpc.py
```diff
@@ -171,9 +171,10 @@ def parse_fetch_response(response: dict):
     return FetchResponse(vectors=vd, namespace=namespace, _check_type=False)
 
 
-def parse_query_response(response: dict, unary_query: bool):
+def parse_query_response(response: dict, unary_query: bool, _check_type: bool = False):
```
miararoy, line 174: I suggest using a similar mechanism as in on the `Index` object where we use kwargs and propagate it this way

pinecone/core/grpc/index_grpc.py
```diff
@@ -171,9 +171,10 @@ def parse_fetch_response(response: dict):
...
     res = []
 
+    # TODO: consider deleting this deprecated case
     for match in response.get('results', []):
         namespace = match.get('namespace', '')
         m = []
```
miararoy, line 180: arent we removing this?

miararoy (changes requested):
aren't we missing the parse_fetch_results as well

pinecone/core/grpc/index_grpc.py
igiloh-pinecone: On the contrary. In `Index` we remove this param from the `kwargs` and enforce our own default value, then explicitly pass it to downstream functions. See for example [here](https://github.com/pinecone-io/pinecone-python-client/blob/main/pinecone/index.py#L97).  
That's why I used the same mechanism here.

(Remember that the top-level function is `query()`, not `parse_response()`).

pinecone/core/grpc/index_grpc.py
igiloh-pinecone: Only if we can remove the `unary_query` param as well, which I wasn't sure about.  
I preferred not to introduce accidentally breaking changes to the coming release.

pinecone/core/grpc/index_grpc.py
benjaminran: Really, I'd actually just hardcode it False w/o enabling users to override; since data is exchanged between client and server as protobuf, the main _check_type usecase of type-checking to detect out-of-sync client & server will already be handled by the protobuf ser/de. Doing _check_type here only checks our grpc->oapi conversion logic, which would be better tested by unit tests. And if we expose the functionality we'll want to document it, and if a user sees that they have an option to "check types" they'll probably do it....

igiloh-pinecone: @benjaminran makes sense.  
Ok, changing it now.

benjaminran (approved):
🤘

harsha2010:
Great work in root causing this!

igiloh-pinecone:
> Great work in root causing this!

@harsha2010 it was actually @benjaminran who pointed us in the right direction based on previous experience.  
I was only verifying the results for our current use case.

pinecone/core/grpc/index_grpc.py
miararoy: 1005

pinecone/core/grpc/index_grpc.py
miararoy: ok, make sense

miararoy (approved):

PR #119, opened 2022-12-29
Batch upsert

acatav:
Implement batch upsert within existing upsert call (only for sync calls)

### DIFF
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -10,6 +10,8 @@ import certifi
 import grpc
 from google.protobuf import json_format
 from grpc._channel import _InactiveRpcError, _MultiThreadedRendezvous
+from tqdm import tqdm
+
 from pinecone import FetchResponse, QueryResponse, ScoredVector, SingleQueryResults, DescribeIndexStatsResponse
 from pinecone.config import Config
 from pinecone.core.client.model.namespace_summary import NamespaceSummary
@@ -255,6 +257,8 @@ class GRPCIndex(GRPCIndexBase):
                vectors: Union[List[GRPCVector], List[Tuple]],
                async_req: bool = False,
                namespace: Optional[str] = None,
+               batch_size: Optional[int] = None,
+               show_progress: bool = True,
                **kwargs) -> Union[UpsertResponse, PineconeGrpcFuture]:
         """
         The upsert operation writes vectors into a namespace.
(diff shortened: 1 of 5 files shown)


tests/unit/test_grpc_index.py
```diff
@@ -75,6 +77,128 @@ def test_upsert_async_upsertInputVectorsAsync(self, mocker):
                 namespace='ns'),
             timeout=None)
 
+    def test_upsert_vectorListIsMultiplyOfBatchSize_vectorsUpsertedInBatches(self, mocker):
+        mocker.patch.object(self.index, '_wrap_grpc_call', autospec=True,
+                            side_effect=lambda stub, upsert_request, timeout: UpsertResponse(
```
igiloh-pinecone, line 82: Nice 😍

igiloh-pinecone (approved):
LGTM!

pinecone/core/grpc/index_grpc.py
```diff
@@ -281,13 +285,21 @@ def upsert(self,
...
+                                  Applied only if batch_size is provided. Default is True.
 
         Returns: UpsertResponse, contains the number of vectors upserted
         """
-
-        args_dict = self._parse_non_empty_args([('namespace', namespace)])
+        if async_req and batch_size is not None:
+            raise ValueError('async_req is not supported when batch_size is provided.'
+                             'To upsert in parallel, please follow: '
+                             'https://docs.pinecone.io/docs/performance-tuning')
```
miararoy, line 302: InvalidArgumentError?

pinecone/core/grpc/index_grpc.py
```diff
@@ -299,12 +311,37 @@ def _vector_transform(item):
...
+        return UpsertResponse(upserted_count=total_upserted)
+
+    def _upsert_batch(self,
+                      vectors: List[GRPCVector],
+                      namespace: Optional[str],
+                      timeout: Optional[float],
+                      **kwargs) -> UpsertResponse:
+        args_dict = self._parse_non_empty_args([('namespace', namespace)])
+        request = UpsertRequest(vectors=vectors, **args_dict)
+        return self._wrap_grpc_call(self.stub.Upsert, request, timeout=timeout, **kwargs)
```
miararoy, line 344: directly return the batch_result.upserted_count will clean up the code on the calling function

pinecone/index.py
```diff
@@ -95,16 +97,51 @@ def upsert(self,
...
+            raise ValueError('async_req is not supported when batch_size is provided.'
+                             'To upsert in parallel, please follow: '
+                             'https://docs.pinecone.io/docs/insert-data#sending-upserts-in-parallel')
+
+        if batch_size is None:
+            return self._upsert_batch(vectors, namespace, _check_type, **kwargs)
+        elif kwargs.get('async_req', False):
+            raise ValueError('async_req is not supported when batch_size is provided.'
+                             'To upsert in parallel, please follow: '
+                             'https://docs.pinecone.io/docs/insert-data#sending-upserts-in-parallel')
```
miararoy, line 121: how would you get here? seems redundant

pinecone/core/grpc/index_grpc.py
acatav: For this version I prefer to use python native exceptions, also it seems like ValueError description fits here well: _"Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError)."_

pinecone/core/grpc/index_grpc.py
acatav: It's a matter of taste, but I prefer it this way for several reasons:

1. This way we can return this result directly when `batch_size` is None, rather then re-create a new response object for no reason
2. Usually when I write functions that do IO I like to have their return type to be the IO result to keep it clean and doing one thing as possible
3. My idea about writing the handle in `batch_size` like this is to have the same code running as we had before, to take as least risk as possible for breaking changes, and here I believe that in the near future 99% of the users will not use the batch size yet

pinecone/index.py
acatav: right, good catch

pinecone/core/grpc/index_grpc.py
miararoy: it makes sesnse for now, but generally, the lower level functions should handle result parsing

miararoy:
LGTM

PR #122, opened 2023-01-02
Fix typo in config.py

eltociear:
explictly -> explicitly


miararoy:
fixed on PR #121 

Thanks for flagging 🙏

PR #134, opened 2023-02-16
Index.upsert() now accepts List[dict]

igiloh-pinecone:
Upsert will now accept vectors represented as dicts. This provides a nice way to pass vectors with sparse values, rather than isntantiating `Vector()` objects

Also, this will allow us to represent entire datasets in the future as dataframes, where each vector is a dictionary (one row of the dataframe).

### DIFF
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -2,6 +2,8 @@
 # Copyright (c) 2020-2021 Pinecone Systems Inc. All right reserved.
 #
 import logging
+import numbers
+import warnings
 from abc import ABC, abstractmethod
 from functools import wraps
 from typing import NamedTuple, Optional, Dict, Iterable, Union, List, Tuple, Any
@@ -25,7 +27,8 @@ from pinecone.core.client.model.sparse_values import SparseValues
 from pinecone.core.grpc.protos.vector_service_pb2_grpc import VectorServiceStub
 from pinecone.core.grpc.retry import RetryOnRpcErrorClientInterceptor, RetryConfig
 from pinecone.core.utils import _generate_request_id, dict_to_proto_struct, fix_tuple_length
-from pinecone.core.utils.constants import MAX_MSG_SIZE, REQUEST_ID, CLIENT_VERSION
+from pinecone.core.utils.constants import MAX_MSG_SIZE, REQUEST_ID, CLIENT_VERSION, REQUIRED_VECTOR_FIELDS, \
+    OPTIONAL_VECTOR_FIELDS
 from pinecone.exceptions import PineconeException
 
 __all__ = ["GRPCIndex", "GRPCVector", "GRPCQueryVector", "GRPCSparseValues"]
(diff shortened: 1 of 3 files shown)


gdj0nes:
Overall I think loosening the restriction is good.

As an aside, would we need to be able to handle numpy arrays to be interoperable with some dataframes?

pinecone/index.py
```diff
@@ -77,18 +77,25 @@ def upsert(self,
...
-            >>>                      values=[1.0, 2.0, 3.0],
-            >>>                      metadata={'key': 'value'}),
-            >>>               Vector(id='id2',
-            >>>                      values=[1.0, 2.0, 3.0],
-            >>>                      sparse_values=SparseValues(indices=[1, 2], values=[0.2, 0.4]))])
+                                     values=[1.0, 2.0, 3.0],
+                                     metadata={'key': 'value'}),
+                              Vector(id='id2',
+                                     values=[1.0, 2.0, 3.0],
+                                     sparse_values=SparseValues(indices=[1, 2], values=[0.2, 0.4]))])
```
gdj0nes, line 93: Is SparseValues necessary here, or can that also be a dict

igiloh-pinecone: @gdj0nes that's coming directly from the `.proto`.  
The proto specifies that `SparseValues` is its own message, not a dict (or proto `Struct`). You have to instantiate a dedicated proto message (which `pb2` exposes as a python class).

If we want `sparse_values` to be a dict or a tuple in the `GRPCVector` message, we need to change the `.proto` (and also the server side) accordingly.

gdj0nes: got it, thanks!

igiloh-pinecone:
@rajat08 @benjaminran I updated the PR and added many more input validations: [diff](https://github.com/pinecone-io/pinecone-python-client/pull/134/commits/5520462569266a15ee19b0c9779818edb40c7233).  

Can you please give the PR another reivew?

pinecone/core/grpc/index_grpc.py
```diff
@@ -317,12 +332,56 @@ def upsert(self,
...
 
+        def _dict_to_grpc_vector(item):
+            item_keys = set(item.keys())
+            if not item_keys.issuperset(REQUIRED_VECTOR_FIELDS):
+                raise ValueError(
+                    f"Vector dictionary is missing required fields: {list(REQUIRED_VECTOR_FIELDS - item_keys)}")
+
+            excessive_keys = item_keys - (REQUIRED_VECTOR_FIELDS | OPTIONAL_VECTOR_FIELDS)
+            if len(excessive_keys) > 0:
+                warnings.warn(f"Found excessive keys in the vector dictionary: {list(excessive_keys)}. "
```
benjaminran, line 343: nit: "Found excess keys"

Also, is this a new usage of the python warnings api? IIRC users need to to programmatically enable "warning output" to see these; we should make sure we have a unified strategy between this and adding warnings as debug logs to the "pinecone.*" logger - and we should probably add some docs to the [Troubleshooting](https://docs.pinecone.io/docs/troubleshooting) page about this

benjaminran: Actually, this inputting vectors as dicts is a net new api in the client, right? It's probably preferable to make this a ValueError and not proceed with the request at all. I'd rather we fail fast even in a few rare cases where a user includes a key which is truly irrelevant rather than silently accepting e.g. upserts with a "metadat" key instead of "metadata" and leaving the user to debug and figure out why their metadata isn't coming back to them on reads.

benjaminran: (This same UX issue is actively biting us in the json api)

rajat08: Agree, keeps it consistent with the tuple variant where we are throwing a value error on more than 3 items

igiloh-pinecone: Totally agree, will change to `raise`.  

BTW `warnings` should be printed (to `stderr`) by default like any other exception (the only difference is they don't stop execution)

igiloh-pinecone:
@benjaminran @rajat08 Fixed all comments. Please approve

rajat08 (approved):

rajat08:
@igiloh-pinecone  , a follow-up to this might be to ensure this new upsert method makes it into the integration tests

acatav (approved):

pinecone/core/grpc/index_grpc.py
```diff
@@ -320,9 +332,20 @@ def upsert(self,
...
+                if len(item) > 3:
+                    raise ValueError(f"Found a tuple of length {len(item)} which is not supported. " 
+                                     f"Vectors can be represented as tuples either the form (id, values, metadata) or (id, values). "
+                                     f"To pass sparse values please use either dicts or a GRPCVector objects as inputs.")
                 id, values, metadata = fix_tuple_length(item, 3)
                 return GRPCVector(id=id, values=values, metadata=dict_to_proto_struct(metadata) or {})
+            elif isinstance(item, Mapping):
+                sparse_values = None
+                if 'sparse_values' in item:
+                    indices = item['sparse_values'].get('indices', None)
```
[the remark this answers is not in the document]
igiloh-pinecone: No, that's actually an allowed (if both `indices` and `values` are `None`).

PR #135, opened 2023-02-20
Use built-in gRPC retry configuration

igiloh-pinecone:
1. Replaced the current implementation of gRPC retries which used an interceptor with a direct config of the gRPC layer itself.
2. Shortened `maxBackoff` from 1.6 sec to 1 sec. When this time is too long, it seems that we're starting to get server-side rejects ("GOAWAY") since the retry messages are still trying to use an old channel.
3. Fixed a bug where `future.result()` for async requests was actually returning `None`s.

### DIFF
--- a/pinecone/core/grpc/index_grpc.py
+++ b/pinecone/core/grpc/index_grpc.py
@@ -11,6 +11,7 @@ import grpc
 from google.protobuf import json_format
 from grpc._channel import _InactiveRpcError, _MultiThreadedRendezvous
 from tqdm import tqdm
+import json
 
 from pinecone import FetchResponse, QueryResponse, ScoredVector, SingleQueryResults, DescribeIndexStatsResponse
 from pinecone.config import Config
@@ -81,8 +82,37 @@ class GRPCIndexBase(ABC):
             "client-version": CLIENT_VERSION
         }
         self._endpoint_override = _endpoint_override
+
+        self.method_config = json.dumps(
+            {
+                "methodConfig": [
+                    {
+                        "name": [{"service": "VectorService.Upsert"}],
+                        "retryPolicy": {
+                            "maxAttempts": 5,
+                            "initialBackoff": "0.1s",
+                            "maxBackoff": "1s",
+                            "backoffMultiplier": 2,
+                            "retryableStatusCodes": ["UNAVAILABLE"],
+                        },
+                    },
+                    {
+                        "name": [{"service": "VectorService"}],
+                        "retryPolicy": {
+                            "maxAttempts": 5,
+                            "initialBackoff": "0.1s",
+                            "maxBackoff": "1s",
+                            "backoffMultiplier": 2,
+                            "retryableStatusCodes": ["UNAVAILABLE"],
+                        },
+                    }
+                ]
+            }
+        )
+
         self._channel = channel or self._gen_channel()
         self.stub = self.stub_class(self._channel)
+
 
     @property
     @abstractmethod
@@ -97,7 +127,9 @@ class GRPCIndexBase(ABC):
         target = self._endpoint()
         default_options = {
             "grpc.max_send_message_length": MAX_MSG_SIZE,
-            "grpc.max_receive_message_length": MAX_MSG_SIZE
+            "grpc.max_receive_message_length": MAX_MSG_SIZE,
+            "grpc.service_config": self.method_config,
+            "grpc.enable_retries": True
         }
         if self.grpc_client_config.secure:
             default_options['grpc.ssl_target_name_override'] = target.split(':')[0]
@@ -111,8 +143,8 @@ class GRPCIndexBase(ABC):
             root_cas = open(certifi.where(), "rb").read()
             tls = grpc.ssl_channel_credentials(root_certificates=root_cas)
             channel = grpc.secure_channel(target, tls, options=_options)
-        interceptor = RetryOnRpcErrorClientInterceptor(self.retry_config)
-        return grpc.intercept_channel(channel, interceptor)
+
+        return channel
 
     @property
     def channel(self):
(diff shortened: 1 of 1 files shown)


rajat08:
@igiloh-pinecone, this is nice!
I did try following some grpc threads about retries a while ago, and I found out that gRPC implements a new version of retires from this release onwards: https://github.com/grpc/grpc/releases/tag/v1.45.0
Maybe we should look into this and see if we can upgrade. Maybe these changes, coupled with upgrades, might give us something more robust. Not too necessary to do at this moment maybe

rajat08:
@igiloh-pinecone  we may wanna bump gRPC version here too to align with retry changes made with newer versions https://github.com/grpc/grpc/releases/tag/v1.45.0

igiloh-pinecone:
> @igiloh-pinecone we may wanna bump gRPC version here too, to align with retry changes made with newer versions https://github.com/grpc/grpc/releases/tag/v1.45.0

Good idea!  
Though I was under the impression that we're keeping the `grpcio` version frozen for the sake of some customers that had too many version conflicts, aren't we?

igiloh-pinecone:
To be perfectly fair - I actually tested this change with the latest `grpcio` version (1.51), not with the one in our requirements.  
I need to make sure it even works with our "supported" grpc version

rajat08:
@igiloh-pinecone  the conflicts mostly arise due to our grpc and proto versions being older than what a user env may have.

rajat08:
@igiloh-pinecone  I dont have date to back this though, it's anecdotal so take it with a grain of salt

rajat08:
> To be perfectly fair - I actually tested this change with the latest `grpcio` version (1.51), not with the one in our requirements. I need to make sure it even works with our "supported" grpc version

okay in that case, this change may not be compatible with versions before 1.45.0 , so let's update

pinecone/core/grpc/index_grpc.py
```diff
@@ -81,9 +82,38 @@ def __init__(self, index_name: str, channel=None, grpc_config: GRPCClientConfig
             "client-version": CLIENT_VERSION
         }
         self._endpoint_override = _endpoint_override
+
+        self.method_config = json.dumps(
+            {
+                "methodConfig": [
+                    {
+                        "name": [{"service": "VectorService.Upsert"}],
```
benjaminran, line 90: Nice catches in this PR :)

Also nice that this gives us the opportunity to differentiate between reads vs writes in our retry strategy; maybe we should consider reducing the read default retries to a max of 3 attempts since they'll generally be more latency sensitive.

But, we should still allow users to override these values like they can today by passing a grpc_config with retry_config set to the GRPCIndex constructor.

miararoy:
@igiloh-pinecone great PR. we should also somehow keep those details in mind for v3

miararoy:
> To be perfectly fair - I actually tested this change with the latest `grpcio` version (1.51), not with the one in our requirements. I need to make sure it even works with our "supported" grpc version

@igiloh-pinecone do we have a way to really test those changes?

igiloh-pinecone:
> > To be perfectly fair - I actually tested this change with the latest `grpcio` version (1.51), not with the one in our requirements. I need to make sure it even works with our "supported" grpc version
> 
> @igiloh-pinecone do we have a way to really test those changes?

Sure, just `pip install` this branch in a clean new environment, then try to upsert 300k vectors in async manner from a poor internet connection.  
I'll try from my home internet soon. Sadly it's poor enough

pinecone/core/grpc/index_grpc.py
igiloh-pinecone: @benjaminran good idea. We'll add that to the task list of Client 3.0.

igiloh-pinecone:
> @igiloh-pinecone we may wanna bump gRPC version here too, to align with retry changes made with newer versions https://github.com/grpc/grpc/releases/tag/v1.45.0

Ok, `1.45` does seem to behave slightly better than `1.44`, so I'm bumping the version.  

But I have to admit I'm not 100% sure this fully solves the problem in case someone tries to upsert hundreds of thousands of vectors all at the same time using `async` calls an internet connection with a very slow upload.   
If too many open requests keep being queued, eventually some of them will get an error. Though maybe that's expected 🤔.

pinecone/core/grpc/index_grpc.py
benjaminran: Will we release this code before client 3.0? We shouldn't break backwards compatibility in the meantime

igiloh-pinecone: Hmmm... I guess you're right.  
I'll add a way to either tweak the params or pass your own `method_config`

benjaminran:
Regarding testing different dependency sets - we used to have unit test support for tox running tests with old dependency versions and more recent ones. Not sure what the state of that is today, but in general that can be a nice thing to dust off for client v3, unless we manage to shed all significant 3p python dependencies. I can track down the PR that added it before if relevant

rajat08:
> 

yeah we will add the tox version tests in v3 back in

rajat08:
> > @igiloh-pinecone we may wanna bump gRPC version here too, to align with retry changes made with newer versions https://github.com/grpc/grpc/releases/tag/v1.45.0
> 
> Ok, `1.45` does seem to behave slightly better than `1.44`, so I'm bumping the version.
> 
> But I have to admit I'm not 100% sure this PR fully solves the problem in case someone tries to upsert hundreds of thousands of vectors all at the same time using `async` calls over an internet connection with a very slow upload. If too many open requests keep being queued, eventually, some of them will get an error. Though maybe that's expected 🤔.

Yeah, and it should error in those cases. I don't think our retries are supposed to help cases like those much, just for intermittent connection stuff

igiloh-pinecone:
> Regarding testing different dependency sets - we used to have unit test support for tox running tests with old dependency versions and more recent ones. Not sure what the state of that is today, but in general that can be a nice thing to dust off for client v3, unless we manage to shed all significant 3p python dependencies. I can track down the PR that added it before if relevant

Client v3.0 will have _zero_ python dependencies :smiley:.  
It's 100% pre-compiled code which is simply bound to python wrapper classes\functions. All the gRPC stuff etc will be built into the wheel.

PR #143, opened 2023-02-22
V2.2.0

igiloh-pinecone:
New release v.2.20


rajat08:
@igiloh-pinecone are these changes covered in docs and integration tests?

rajat08:
> @igiloh-pinecone are these changes covered in docs and integration tests?

additionally , are protos in the protos repo up to date?

igiloh-pinecone:
> @igiloh-pinecone are these changes covered in docs and integration tests?

There are UTs in the repo for the recent changes, no new integ tests.  
We have tested the functionality manually against the beta env. We will need to add integ tests as soon as possible.

As for docs - they are documented in the docstrings of the functions themselves, and in the CHANGELOG.

rajat08:
__generally__ we should have all these changes in before we do pypi releases, but it's okay i guess if it was urgent

rajat08:
We have release out in pypi that is from code that's merged yet. We should avoid doing this

igiloh-pinecone:
Anyway I suggest we merge this to `main` now since it's de-facto released anyway.

All the code changes were reviewed and approved in previous individual PRs, and unit tests are passing.  
We can release another minor version with documentation updates in a few days, if we find that anything is missing.

rajat08 (dismissed):

pinecone/index.py
```diff
@@ -151,12 +167,65 @@ def _upsert_batch(self,
 
         args_dict = self._parse_non_empty_args([('namespace', namespace)])
```
rajat08, line 168: nit: this is not needed for upsert, namespace is the only arg

pinecone/index.py
```diff
@@ -169,6 +241,37 @@ def _vector_transform(item: Union[Vector, Tuple]):
             **{k: v for k, v in kwargs.items() if k in _OPENAPI_ENDPOINT_PARAMS}
         )
 
+    def upsert_from_dataframe(self,
+                         df,
+                         namespace: str = None,
+                         batch_size: int = 500,
```
rajat08, line 247: default batch size is too high

pinecone/index.py
```diff
@@ -169,6 +241,37 @@ def _vector_transform(item: Union[Vector, Tuple]):
...
+                         show_progress: bool = True) -> None:
+        """Upserts a dataframe into the index.
+
+        Args:
+            df: A pandas dataframe with the following columns: id, vector, and metadata.
```
rajat08, line 252: why does it not support sparse values? If it does we need to update this doc string

pinecone/index.py
```diff
@@ -169,6 +241,37 @@ def _vector_transform(item: Union[Vector, Tuple]):
...
+            namespace: The namespace to upsert into.
```
rajat08, line 253: Why not have namespace in the schema rather than a explicit arg here

rajat08:
This PR adds numpy, tqdm and pandas dependencies, and I don't see too much value putting it in the client. I am not sure why we are adding them, but I'll defer to product cc @gdj0nes

rajat08:
Nit : In index.py we have delete return `Dict[str, Any] ` , the delete call from the backend sends us nothing but whether the call was accepted, we should stick to that

rajat08:
Nit: the additional args check that we do for `upsert` should be done for all data plane calls

pinecone/index.py
```diff
@@ -151,12 +167,68 @@ def _upsert_batch(self,
 
         args_dict = self._parse_non_empty_args([('namespace', namespace)])
```
rajat08, line 168: nit : I dont think the parse_non_empty_args function is too necessary, we are not doing any extra validation there

rajat08:
I think we have some really nice tests in the repo now. Let's try to make them part of the CI, use tox, fixtures etc.

rajat08:
@igiloh-pinecone  I believe we still need to update grpc version right

pinecone/index.py
```diff
@@ -169,6 +241,37 @@ def _vector_transform(item: Union[Vector, Tuple]):
             **{k: v for k, v in kwargs.items() if k in _OPENAPI_ENDPOINT_PARAMS}
         )
 
+    def upsert_from_dataframe(self,
+                         df,
+                         namespace: str = None,
+                         batch_size: int = 500,
+                         show_progress: bool = True) -> None:
```
rajat08, line 248: We should set this to `false` by default we don't want to print out stuff to users stdout without them expecting it to

pinecone/core/grpc/index_grpc.py
```diff
@@ -274,18 +310,25 @@ def upsert(self,
...
         Examples:
-            >>> index.upsert([('id1', [1.0, 2.0, 3.0], {'key': 'value'}), ('id2', [1.0, 2.0, 3.0])],
-            >>>               namespace='ns1', async_req=True)
+            >>> index.upsert([('id1', [1.0, 2.0, 3.0], {'key': 'value'}),
+                              ('id2', [1.0, 2.0, 3.0])
+                              ],
+                              namespace='ns1', async_req=True)
+            >>> index.upsert([{'id': 'id1', 'values': [1.0, 2.0, 3.0], 'metadata': {'key': 'value'}},
+                              {'id': 'id2',
+                                        'values': [1.0, 2.0, 3.0],
```
rajat08, line 319: nit : values twice

pinecone/index.py
rajat08: Do we have docs for these interface changes?

rajat08:
My 2 cents: We added tqdm, pandas, and numpy dependencies here that give us little value. I don't see why we need to batch for the users; a user can shoot themselves in the foot here; I lean towards keeping APIs simple unless we do something significantly optimized behind the scenes. Just putting it out there, I would leave the final call to others

pinecone/core/grpc/index_grpc.py
```diff
@@ -359,6 +450,52 @@ def _upsert_batch(self,
...
+            results = [async_result.result() for async_result in tqdm(results, desc="collecting async responses")]
+
+        upserted_count = 0
+        for res in results:
+            upserted_count += res.upserted_count
+
+        return UpsertResponse(upserted_count=upserted_count)
+
+    @staticmethod
+    def _iter_dataframe(df, batch_size):
```
rajat08, line 494: How does this compare to using `groupBy` and `apply` using an upsert function to batches of rows in the df?

pinecone/index.py
```diff
@@ -169,6 +241,37 @@ def _vector_transform(item: Union[Vector, Tuple]):
...
+            import pandas as pd
+        except ImportError:
+            raise RuntimeError("The `pandas` package is not installed. Please install pandas to use `upsert_from_dataframe()`")
+
+        if not isinstance(df, pd.DataFrame):
+            raise ValueError(f"Only pandas dataframes are supported. Found: {type(df)}")
+
+        upserted_count = 0
+        pbar = tqdm(total=len(df), disable=not show_progress)
+        for i in range(0, len(df), batch_size):
```
rajat08, line 267: this might be overoptimizing but the `itertools.islice` method that we generally use in examples can be more efficient

miararoy (approved):

PR #168, opened 2023-04-28
Backfill missing unit tests for pinecone.init() method

jhamon:
### Description

A lot of refactoring is needed in `init()` method to remove various UX footguns, improve performance (e.g. whoami), and give better error messages when the client is misconfigured. But before we can safely do any of that, we need to add a lot of missing tests to document the existing behavior of the `init()` method.

### Asana Ticket

https://app.asana.com/0/1203260648987893/1204499608161625/f

### Type
- [ ] Bug fix
- [ ] New feature
- [ ] Infrastructure change
- [ ] Non-code (docs/comments/lint etc.)
- [x] Something else: Tech debt / missing tests

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -1,14 +1,211 @@
 import pinecone
 from pinecone.config import Config
+from pinecone.core.client.configuration import Configuration as OpenApiConfiguration
 
+import pytest
+import tempfile
+import os
 
-def _test_multi_init():
-    env = 'test-env'
-    api_key = 'foobar'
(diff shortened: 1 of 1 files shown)


jhamon:
When developing this locally I was just running `pytest --maxfail=1 tests/unit` to run tests. Looking at this github actions output, I'm not confident tox is actually doing anything inside each of these matrixed test runs. Not familiar with tox, need to dig in more to understand what's going on in there. 🤔

micmmakarov (approved):

PR #169, opened 2023-04-28
Fix tox configs for CI testing

jhamon:
## Problem

I discovered recently that tests have not been running in CI because some configuration for our test orchestrator [`tox`](https://tox.wiki/en/latest/config.html) were [accidentally removed](https://github.com/pinecone-io/pinecone-python-client/commit/10fb42c2983b731700ffe89183994cd4e1844020). This fact was obscured because the CI configuration specifying a matrix of python versions to test meant several versions of the workflow were kicking off, but if you look carefully you'd see no tests were actually executing.

## Solution

1. Add and iterate on configuration to `tox.ini`, the configuration for the tox test orchestrator. I'm new to this tool, but it seems that tox works in conjunction with [tox-gh-actions](https://github.com/ymyzk/tox-gh-actions) and the github matrix strategy to execute tests on multiple versions of python. We have to declare each test environment, and map those names to `gh-actions` names for these environments in an additional configuration section in the `tox.ini` file.

2. Make some adjustments to the "old deps" versions in order to get the python 3.7 installation & tests to run.

3. Fix tests that are revealed to be failing once all tests are running in CI. The changes are fairly minor, basically splitting up one of the parameterized tests so I could make a small adjustment to the assertion being made when invalid vector is are passed to upsert.

## Test Plan

Look closely at the checks that are running, and observe that test output shows the builds actually ran tests.

### Other changes and notes

- Bumped us up to `actions/setup-python@v4` from `actions/setup-python@v2` to fix a deprecation error message.
- I omitted Python 3.6 from the test matrix since it was end-of-life in 2021 and is no longer supported by github actions for testing on `ubuntu-latest`. A [relevant comment](https://github.com/actions/setup-python/issues/544#issuecomment-1332535877).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -7,7 +7,6 @@ jobs:
     name: Run tests
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
         python-version: [3.7, 3.8, 3.9, '3.10', 3.11]
 
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -7,6 +7,7 @@ jobs:
     name: Run tests
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: [3.7, 3.8, 3.9, '3.10', 3.11]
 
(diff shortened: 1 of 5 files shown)


igiloh-pinecone (approved):
LGTM

PR #178, opened 2023-06-03
Release from CI

jhamon:
## Problem

A good release process needs to be fast while also ensuring quality and consistency. Confidence of quality comes from incorporating automated testing. And there are many aspects of consistency including:

- Ensuring commits are tagged with version number so we can know what is in each release (now and in the future in case of any security reports)
- Version number in `pinecone/__version__` is correctly updated, so we don't try to reuse the same one later
- Building in a standardized way to avoid introducing accidental inconsistencies (e.g. uncommitted and untested local changes accidentally included in a release, using an old version of python on a local machine, etc)

## Solution

Implement github actions workflows for publishing releases (`release.yaml`) and prereleases (`alpha-release.yaml`).

Screenshot 2023-06-06 at 3 01 23 PM

These workflows both follow the same high-level steps that take place in a `publish-to-pypi.yaml` [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows):
- Run tests using another new reusable workflow, `testing.yaml`, which I refactored out of the existing PR workflow.
- Automatic version bumps based on type of release (`major`, `minor`, `patch`). 
  - I ended up writing my own small action to do this because I didn't find a nice github action that would do this off the shelf. There are lots of small code packages out there that do the version bump logic but they tended to have weird flaws that made them a poor fit (e.g. wanting to handle the git commit part even though this isn't the right moment for it in my workflow, wanting to store separate version-related config elsewhere, poor support for custom alpha/rc version numbers, etc.)
- Check that version numbers are unique (i.e. that there is no preexisting git tag with that number. This mainly protects us from accidentally reusing the same rc number). This validation is possible since git tags become the source of truth on which version numbers are used/available.
- Build with existing make task
- Upload to PyPI using twine. This is no change, just using Makefile tasks defined by others in the past.
- Git push commit with updated `pinecone/__version__` and git tags. Save this step for last so that the git tag is not consumed unless the pypi upload is successful.

## Other notes

Updated the version saved in `pinecone/__version__` because I noticed that it is out of date with the latest version published on PyPI. This is one example of a type of error that will no longer occur when publishing from CI.

While implementing this I ran into several different issues in github actions including: 

- https://github.com/actions/runner/issues/2472
- https://github.com/actions/runner/issues/1483
- https://github.com/actions/hello-world-javascript-action/issues/12

## Future Work

Some ideas for future: 
- I would like to make some automated changelog updates and draft release notes from CI. But not doing this now as this is already a large change.
- I could also do some work to figure out what the automatic next rc number is by inspecting what git tags have been used. But for now it's easy to just pass in the values as an input when starting the job.

## Type of Change

## Test Plan

I used the test environment at test.pypi.org to publish an rc version of this package using the `alpha-release.yaml` version of the workflow. 

Here's a [test run](https://github.com/pinecone-io/pinecone-python-client/actions/runs/5192266625) I did that targeted the testpypi and resulted in [this 2.2.2rc4 artifact](https://test.pypi.org/project/pinecone-client/2.2.2rc4/). Which can be installed like this by specifying the test.pypi.org index:

```
pip install -i https://test.pypi.org/simple/ pinecone-client==2.2.2rc4
```

In general, to test a github workflow that only exists in a branch you need to install the github CLI and run a command like this from inside the project:

```
gh workflow run alpha-release.yaml -f ref=jhamon/release-from-ci -f releaseLevel=patch -f prereleaseSuffix='rc4'
```

The only thing I changed after this run was to switch from targeting the test PyPI index to the production PyPI index.

The prod workflow is substantially similar so I have a high confidence it should work but I don't want to test it and release until I land this in master because I want to the release commit to tag a SHA in the main history (and not this branch). I plan to land, then make a follow up for any small issues that come up when trying to use it for the first time.


pinecone/__version__
```diff
@@ -1 +1 @@
-2.2.0
```
jhamon, line 1: A past error meant this version number was out of sink with what has already been published to pypi. So needed to update that for future releases.

.github/actions/bump-version/action.js
```diff
@@ -0,0 +1,38 @@
...
+  core.setOutput('version_tag', `v${newVersion}`)
+
+  return newVersion;
+}
+
+function calculateNewVersion(currentVersion, bumpType) {
+  const [major, minor, patch] = currentVersion.split('.');
+  let newVersion;
+
+  switch (bumpType) {
```
damargulis, line 21: I know you said you looked at a few actions (I remember having the same struggle lol), but I eventually got https://github.com/mathieudutour/github-tag-action to work for our use case in the dashboard repo -- if you use that with dry_run, does that work for you here?

.github/actions/bump-version/action.js
```diff
@@ -0,0 +1,38 @@
+const core = require('./core');
+
+function bumpVersion(currentVersion, bumpType, prerelease) {
+  let newVersion = calculateNewVersion(currentVersion, bumpType);
+
+  if (prerelease !== undefined && prerelease !== '') {
```
damargulis, line 6: assuming prerelease is a string, you can also just do `if(prerelease)`

.github/actions/bump-version/core.js
```diff
@@ -0,0 +1,79 @@
+// Copied these commands out of the github actions toolkit
+// because actually depending on @actions/core requires me to check
+// in node_modules and 34MB of dependencies, which I don't want to do.
```
damargulis, line 3: You should be able to run npm install as part of the workflow to avoid checking it in

damargulis (approved):

.github/actions/bump-version/action.js
jhamon: That looks like a nice action overall, but it presumes you are trying to use [semantic release](https://github.com/semantic-release/semantic-release) commit message conventions. Semantic release is a nice idea we should consider adopting in our open source projects in the future but is not something we are doing currently. This action lets you specify the "default bump" which would give the behavior I want now, but I don't really want this action to make a different determination about release version than the one I ask for somewhere down the road because of how somebody might format a commit message.

.github/actions/bump-version/core.js
jhamon: That said, I can understand why they would want to encourage checking in node_modules for shared actions even though it seems gross to js people. Checking in allows actions to run predictably and consistently every time without things going wrong due to npm.org being offline or loose versioning (e.g. with >= instead of exact pinning) leading to unexpected changes of behavior in a future install. My issue really wasn't with committing dependencies as with the absolutely huge amount of dependencies being pulled in when all I wanted was to read and write a few environment variables.

PR #190, opened 2023-07-09
Create method in Index class for fetching IDs of all vectors in index

Tylersuard:
## Problem
Pinecone does not have any way for users to fetch all IDs of all vectors in an index.

## Solution
Loop and query until the number of unique IDs matches the number of vectors in the index.

## Type of Change

- [ X] Bug fix (non-breaking change which fixes an issue)
- [ X] New feature (non-breaking change which adds functionality)
- [X ] This change requires a documentation update

## Test Plan
Try it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,19 @@ query_response = index.query(
 )
 ```
 
+## Get All Vector IDs
+
+The following example retrieves the ID for every vector in the Index.
+```python
+import pinecone
+
+pinecone.init(api_key="YOUR_API_KEY", environment="us-west1-gcp")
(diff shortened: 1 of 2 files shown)


Tylersuard:
@rajat08 @benjaminran

daverigby:
Note that v3.1 has [introduced the `list` and `list_pagnated` API methods](https://github.com/pinecone-io/pinecone-python-client/releases/tag/v3.1.0) (Serverless only) which use a server-side implementation which should be significantly quicker than the brute-force search approach used above.

PR #195, opened 2023-08-19
Replace deprecated HTTPResponse.getheader()

izeye:
## Problem

See gh-125

## Solution

This PR simply replaces the deprecated `HTTPResponse.getheader()` usages with the suggested `HTTResponse.headers`.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Infrastructure change (CI configs, etc)
- [ ] Non-code change (docs, etc)
- [x] None of the above: Replacing deprecated usages

## Test Plan

Describe specific steps for validating this change.

PR #197, opened 2023-08-24
update urllib3_response.headers.get as getheader is deprecated

tdonia:
## Problem

Follow up to #195, which updated to use `urllib.headers` -- there's also a deprecation for `getheader` coming: https://github.com/urllib3/urllib3/pull/2814

## Solution

Updated the call in `rest.py` to use `.get` instead of `.getheader`; based on the [code](https://github.com/urllib3/urllib3/blob/1.2.1/urllib3/response.py#L191C29-L191C32) this should be compatible with 1.2.1, [our current min version](https://github.com/pinecone-io/pinecone-python-client/blob/main/requirements.txt#L8)

## Type of Change

## Test Plan

REST requests should work without a notice in URLLib3 2.0.4+

### DIFF
--- a/pinecone/core/client/rest.py
+++ b/pinecone/core/client/rest.py
@@ -42,7 +42,7 @@ class RESTResponse(io.IOBase):
 
     def getheader(self, name, default=None):
         """Returns a given response header."""
-        return self.urllib3_response.headers.getheader(name, default)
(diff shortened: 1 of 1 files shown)


jhamon (approved):
Thanks for the attention to detail on this 🚢

Issue #196, opened 2023-08-20
[Bug] Python client connects to wrong environment 


Amar-89:
### Is this a new bug in the Pinecone Python client?

- [X] I believe this is a new bug in the Pinecone Python Client
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Python client automatically tries to connect to "us-west1-gcp" even though the env variable is set to "asia-southeast1-gcp-free". 

`PINECONE_ENV=asia-southeast1-gcp-free`

```
pinecone.core.client.exceptions.UnauthorizedException: (401)
Reason: Unauthorized
HTTP response headers: HTTPHeaderDict({'www-authenticate': 'API key is missing or invalid for the environment "us-west1-gcp". Check that the correct environment is specified.', 'content-length': '114', 'date': 'Sun, 20 Aug 2023 17:10:54 GMT', 'server': 'envoy'})
HTTP response body: API key is missing or invalid for the environment "us-west1-gcp". Check that the correct environment is specified.
```

### Expected Behavior

It should connect to the southeast asia server instead of trying to connect to US-west.

### Steps To Reproduce

Using poetry with openai, pinecone, langchain.

### Environment

```markdown
- OS: Ubuntu
- Python: 3.11.4
- pinecone: n/a
```

jhamon:
You are setting the incorrect environment variable name and getting the default environment as a result. You should set `PINECONE_ENVIRONMENT=asia-southeast1-gcp-free` instead of `PINECONE_ENV`. We have a ticket to improve error messaging in cases like this.

PR #198, opened 2023-08-24
adds util for checking kwargs, and uses it to check init()

tdonia:
## Problem

Per #196 a typo in init's kwargs results in a default that might not be expected.

## Solution

Utils now includes check_kwargs, which takes a caller + the given kwargs, and fails with a traceback if they don't match.

## Open Questions
cc @jhamon @austin-denoble @rohanshah18 
1) Where else should we use check_kwargs?
2) Is the traceback useful? in this instance it risks leaking api keys into logs, so for `init` in particular i question its value.

## Type of Change

## Test Plan

`pinecone.init(not_a_kwarg='foo')` should fail

### DIFF
--- a/pinecone/core/utils/__init__.py
+++ b/pinecone/core/utils/__init__.py
@@ -109,3 +109,10 @@ def load_strings_public(proto_arr: 'vector_column_service_pb2.NdArray') -> List[
 def warn_deprecated(description: str = '', deprecated_in: str = None, removal_in: str = None):
     message = f'DEPRECATED since v{deprecated_in} [Will be removed in v{removal_in}]: {description}'
     warnings.warn(message, FutureWarning)
+
+def check_kwargs(caller, given):
+    import inspect
+    argspec = inspect.getfullargspec(caller)
+    diff = set(given).difference(argspec.args)
+    if diff:
+        raise TypeError(caller.__name__ + ' had unexpected keyword argument(s): ' + ', '.join(diff))
--- a/pinecone/core/utils/__init__.py
+++ b/pinecone/core/utils/__init__.py
@@ -1,6 +1,8 @@
 #
 # Copyright (c) 2020-2021 Pinecone Systems Inc. All right reserved.
 #
+import inspect
+import logging
 import re
 import uuid
 import warnings
@@ -111,8 +113,7 @@ def warn_deprecated(description: str = '', deprecated_in: str = None, removal_in
     warnings.warn(message, FutureWarning)
 
 def check_kwargs(caller, given):
-    import inspect
     argspec = inspect.getfullargspec(caller)
     diff = set(given).difference(argspec.args)
     if diff:
-        raise TypeError(caller.__name__ + ' had unexpected keyword argument(s): ' + ', '.join(diff))
+        logging.exception(caller.__name__ + ' had unexpected keyword argument(s): ' + ', '.join(diff), exc_info=False)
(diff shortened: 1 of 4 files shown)


tdonia:
here's example output with the traceback -- note the `xxxx-xxxx-xxxx-xxxx-xxxx` would be sensitive, and ideally not logged:

```
(popy-py3.11) popy :> poetry run python popy/py-cone.py
Traceback (most recent call last):
  File "/Users/travisdonia/src/popy/popy/py-cone.py", line 6, in <module>
    pinecone.init(wrong_kwarg="xxxx-xxxx-xxxx-xxxx-xxxx", environment="northamerica-northeast1-gcp")
  File "/Users/travisdonia/Library/Caches/pypoetry/virtualenvs/popy-b0Yl8H78-py3.11/lib/python3.11/site-packages/pinecone/config.py", line 248, in init
    check_kwargs(init, kwargs)
  File "/Users/travisdonia/Library/Caches/pypoetry/virtualenvs/popy-b0Yl8H78-py3.11/lib/python3.11/site-packages/pinecone/core/utils/__init__.py", line 118, in check_kwargs
    raise TypeError(caller.__name__ + ' had unexpected keyword argument(s): ' + ', '.join(diff))
TypeError: init had unexpected keyword argument(s): wrong_kwarg
```

pinecone/core/utils/__init__.py
```diff
@@ -109,3 +109,10 @@ def load_strings_public(proto_arr: 'vector_column_service_pb2.NdArray') -> List[
 def warn_deprecated(description: str = '', deprecated_in: str = None, removal_in: str = None):
     message = f'DEPRECATED since v{deprecated_in} [Will be removed in v{removal_in}]: {description}'
     warnings.warn(message, FutureWarning)
+
+def check_kwargs(caller, given):
+    import inspect
```
jhamon, line 114: I don't think it's common practice to have an import statement inside of a function definition. Seems like this could have some performance implication if we start using `check_kwargs` in more places, particularly on data plane operations which may be high volume. I would move this import to the top of the file.

And I know there's a ton of untested code in this project, but it would be great to see a test case or two showing what this does in different scenarios.

jhamon (approved):
I see what you mean about the problem of sensitive information in the stacktrace. I found this old package, which is currently archived, which [implemented a decorator-based approach](https://github.com/untitaker/python-sensitive-variables/tree/master) to scrubbing sensitive information from stacktraces. Maybe you could look around to find a similar project that isn't dead or else lift the relevant bits of source for us to directly copy. It doesn't look like that much code, and it's MIT licensed so fair game.

I think we can still move forward with this change because the problem of secrets in stacktraces is not a new issue. Any kind of error that results in a thrown exception would already have this problem if the person is initializing with method params (instead of environment variables, which should not appear). And stacktrace secret-scrubbing could be handled as a follow-up item.

As for where else this util should get used... I'm not sure. Most of the bug reports I have seen along these lines were problems with the init call. I think once people get started, they're more likely to take an iterative approach if something isn't working  the way they expect due to them passing the wrong stuff. But when they run into this on their very first line of code with Pinecone, they seem to just throw up their hands in frustration.

tdonia:
I've modified this specifically to omit the traceback for this particular check because the risk of leaking an API Key in a bad `init` call is especially high. also added a test and moved that import up.  thanks for the review!

rohanshah18 (approved):
LGTM!

PR #204, opened 2023-09-27
Apache license 2.0

gdj0nes:
Update license to Apache 2.0

## Type of Change


setup.py
```diff
@@ -21,7 +19,7 @@ def read(fname):
     name="pinecone-client",
     version=read("pinecone/__version__").strip(),
     description="Pinecone client and SDK",
-    license="Proprietary License",
+    license='Apache 2.0',
```
jhamon, line 22: Seems like you need to run black again. It favors double quotes.

jhamon (approved):
Looks good 🚢

setup.py
gdj0nes: i don't think black catches the setup.py might need to tweek that

PR #212, opened 2023-10-04
Add new setup-poetry action and fix nightly-release

austin-denoble:
## Problem
`nightly-release` workflow is still failing after the previous "fix": https://github.com/pinecone-io/pinecone-python-client/pull/208.

This time it's due to `make package` failing on `poetry build`, because `poetry` wasn't setup properly in the workflow.
https://github.com/pinecone-io/pinecone-python-client/actions/runs/6399744230/job/17372332268

> Run make package
poetry build
make: poetry: No such file or directory
make: *** [Makefile:22: package] Error [12](https://github.com/pinecone-io/pinecone-python-client/actions/runs/6399744230/job/17372332268#step:9:13)7
Error: Process completed with exit code 2.

While fixing this I noticed we're also not installing Poetry in the `publish-to-pypi` workflow, so calling `poetry version` and `make` commands there would fail. We need to make sure we're installing Poetry properly before using `make` in our workflows, as they call Poetry under the hood.

## Solution
- Add a new `setup-poetry` action that encapsulates the install + dependencies bits. I noticed we were doing this explicitly in several workflows and thought it called for an action.
- Fix `nightly-release` worfklow: call the new action before calling `make package`.
- Fix `release-to-pypi` worfklow: call the new action before calling `poetry version`, and `make` commands for publishing.
- Clean up / use the action in `testing` worfklow.

## Type of Change

## Test Plan
Validate PR workflows pass as expected. Merge and check `nightly-release` to make sure we're in good shape.

### DIFF
--- a/.github/workflows/publish-to-pypi.yaml
+++ b/.github/workflows/publish-to-pypi.yaml
@@ -12,7 +12,7 @@ on:
         description: 'Git ref to build (branch name or SHA)'
         required: true
         type: string
-        default: 'main' 
+        default: 'main'
       releaseLevel:
         description: 'Release level'
         required: true
@@ -59,10 +59,10 @@ jobs:
         id: bump
         uses: './.github/actions/bump-version'
         with:
-          versionFile: pinecone/__version__ 
+          versionFile: pinecone/__version__
           bumpType: ${{ inputs.releaseLevel }}
           prereleaseSuffix: ${{ inputs.prereleaseSuffix }}
-      
+
       - name: Verify unique release number
         run: |
           TAG_NAME=${{ steps.bump.outputs.VERSION_TAG }}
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-to-pypi.yaml
```diff
@@ -71,19 +71,23 @@ jobs:
...
+          python-version: 3.x
+
+      - name: Setup Poetry
+        uses: ./.github/actions/setup-poetry
 
       - name: Set up Git
         run: |
           git config --global user.name "Pinecone CI"
           git config --global user.email "clients@pinecone.io"
-
```
austin-denoble, line 82: Moved this up, did a tiny bit of organizing steps in here.

.github/workflows/testing.yaml
```diff
@@ -12,37 +12,21 @@ jobs:
...
-      run: make tests
-
-    - name: Build Python client
-      run: make package
-
-    - name: Build docs
-      run: make docs
-
-    - name: Build Project
-      run: poetry build
```
austin-denoble, line 48: Removed, we're calling `make package` above which is essentially an alias for `poetry build`, don't think we need to call it again.

jhamon (approved):
Nice work 🚀

PR #215, opened 2023-10-11
Update generated client code

jhamon:
## Problem

We need to update the generated core module to reflect upcoming API changes

## Solution

- Migrate from v5 to v7 openapi generator to avoid generating code with deprecated urllib3 usage.
- Regenerate contents of `pinecone.core` using pinecone-protos (private) repo.
- Refactor to move all non-generated items out from under `pinecone.core`. These never should have been under `pinecone.core` in the first place.
  - Added `pinecone.utils`
  - Added `pinecone.grpc` to hold `GRPCIndex` and related code.
- Add new dependencies required by updated generated code (`pydantic` for typechecking, `aenum` for enums)
- Disable legacy `integ` tests which won't work until we refactor configuration.

## Type of Change


pyproject.toml
```diff
@@ -57,18 +57,20 @@ classifiers=[
...
+urllib3 = "1.25.3"
 tqdm = ">=4.64.1"
 numpy = ">=1.22.0"
 grpcio = ">=1.44.0"
 grpc-gateway-protoc-gen-openapiv2 = "0.1.0"
 googleapis-common-protos = ">=1.53.0"
 lz4 = ">=3.1.3"
 protobuf = "~=3.20.0"
+aenum = "3.1.11"
+pydantic = "1.10.12"
```
jhamon, line 73: These are new dependencies required by the openapi generator. Both have permissive open source licenses, so no issue there.

jhamon, line 73: Updated and added dependencies  to satisfy the requirements of the generated openapi code. Both have permissive open source licenses, so no issue there.

tests/integ/test_index.py
```diff
@@ -8,128 +8,131 @@
 from pinecone import ApiTypeError, ApiException
 
 responses_req = Responses()
-responses = Responses('requests.packages.urllib3')
+responses = Responses("requests.packages.urllib3")
 
+@pytest.mark.skip()
```
jhamon, line 13: There are a lot of black changes in this file, but all the tests are currently marked `skip` because they can't pass until we overhaul configuration.

The urllib3_mock we are depending on throws a lot of deprecation warnings and appears to be abandoned so we should figure out some other tool there if we need to do request mocking in the future. But in reality I don't think we need to mock requests at that level if we refactor our tests into either pure unit tests or real integration tests against a live environment.

tests/unit/test_config.py
```diff
@@ -6,210 +6,204 @@
 import tempfile
 import os
 
+
 @pytest.fixture(autouse=True)
```
jhamon, line 10: I removed references to project name from these config tests.

tests/unit/test_grpc_index.py
jhamon, line 1: I don't think much changed inhere besides formatting.

tests/unit/test_index.py
```diff
@@ -1,437 +1,464 @@
 import pandas as pd
-import numpy as np
 import pytest
 import warnings
 
-from pinecone.core.client.api_client import Endpoint
```
jhamon, line 6: This export is not available, and most substantive changes in this test file are to refactor away from depending on it.

.github/actions/bump-version/action.js
```diff
@@ -6,10 +6,10 @@ function bumpVersion(currentVersion, bumpType, prerelease) {
   if (prerelease) {
     newVersion = `${newVersion}.${prerelease}`;
   }
-  core.setOutput('previous_version', currentVersion)
-  core.setOutput('previous_version_tag', `v${currentVersion}`)
+  core.setOutput('previous_version', currentVersion);
```
jhamon, line 9: Maybe black did this? Not really my intention, sorry.

pyproject.toml
austin-denoble: Could potentially use pydantic in a similar way to typebox in the TS client if we wanted to add better runtime validation and error handling, if I'm understanding things correctly.

tests/unit/test_index.py
austin-denoble: Good call, thank you.

tests/unit/test_index.py
```diff
@@ -1,437 +1,464 @@
...
-        self.sv2 = {
-            'indices': self.svi2,
-            'values': self.svv2
-        }
+        self.sv2 = {"indices": self.svi2, "values": self.svv2}
 
-        pinecone.init(api_key='example-key')
-        self.index = pinecone.Index('example-name')
-    
-    def test_upsert_numpy_deprecation_warning(self, mocker):
```
austin-denoble, line 40: **question:** Are we still planning on deprecating numpy? I've seen references to that in the codebase and this seems to confirm that. 

If so we can probably add a ticket.

pinecone/grpc/index_grpc.py
```diff
@@ -15,8 +15,8 @@
```
austin-denoble, line 15: This will be good for actually adding some documentation around the GRPC bits, thank you. 👍

austin-denoble (approved):
Thank you for all the work that went into getting the core regenerated and updated. Overall this looks good, it seems like we'll need a few tickets for follow-up here around re-enabling integration tests, and replacing `urllib3` if needed for mocking.

It looks like the pdoc build is also failing due to a couple import related issues. I think this is due to it using dynamic analysis when extracting docstrings, so Python modules are executed/imported when it runs:
- `ModuleNotFoundError: No module named 'pinecone.core.client.model'`
- `ImportError: cannot import name 'Endpoint' from 'pinecone.core.client.api_client' (/home/runner/work/pinecone-python-client/pinecone-python-client/pinecone/core/client/api_client.py)`

I can pull things down and maybe take a look, we could also skip / disable for now since we're in a feature branch. I think this also still may need a rebase against `origin/spruce` to pull in the workflow changes.

tests/unit/test_index.py
jhamon: Yeah, we should remove it before releasing spruce. I'll add a ticket.

PR #219, opened 2023-10-17
Reformat docstrings across Python client

austin-denoble:
## Problem
In a previous revision we added CI tooling for generating and deploying documentation for the Python client. https://github.com/pinecone-io/pinecone-python-client/pull/207

There was a lot of follow-up needed to clean up the generated documentation. A lot of our existing docstrings were also of disparate formatting types.

## Solution
- Exclude the `/pinecone/core/*` modules and `info.py` module from the built output.
- Align our manually maintained docstrings around the Google format: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings.
- Add examples, fix up wording, flesh out descriptions, etc.

It was tricky aligning on a specific docstring format to go with. It seems like a lot of the auto-generated comments in `/core/` are of either reST (reStructuredText, the sphinx default) or Google. A lot of the docstrings that were already stubbed out were of the Google style.

Since pdoc handles both types I went with Google as it felt a bit terser in terms of number of lines. Ultimately, I think most docs utilities offer plugins or options to handle either type.

## Type of Change
- [X] Non-code change (docs, etc)

## Test Plan
Easiest way to validate the output would probably be to pull this branch down locally and run the pdoc utility:

```
$ poetry shell
$ poetry install

# same command we use in the `build-docs` action
$ poetry run pdoc pinecone/ '!pinecone.core' '!pinecone.info' --favicon ./favicon-32x32.png --docformat google -o ./docs
```

Once built, you can open the `index.html` file within the `/docs/` folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-# pinecone-client
+# pinecone-python-client
+
 The Pinecone python client
 
 For more information, see the docs at https://www.pinecone.io/docs/
@@ -6,6 +7,7 @@ For more information, see the docs at https://www.pinecone.io/docs/
 ## Installation
 
 Install a released version from pip:
+
 ```shell
 pip3 install pinecone-client
 ```
@@ -17,11 +19,13 @@ pip3 install "pinecone-client[grpc]"
 ```
 
 Or the latest development version:
+
 ```shell
 pip3 install git+https://git@github.com/pinecone-io/pinecone-python-client.git
 ```
 
 Or a specific development version:
+
 ```shell
 pip3 install git+https://git@github.com/pinecone-io/pinecone-python-client.git
 pip3 install git+https://git@github.com/pinecone-io/pinecone-python-client.git@example-branch-name
@@ -120,7 +124,6 @@ index = pinecone.Index("example-index")
 index_stats_response = index.describe_index_stats()
 ```
 
-
 ## Upsert vectors
 
 The following example upserts vectors to `example-index`.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,11 +1,13 @@
-# pinecone-client
+# pinecone-python-client
```
austin-denoble, line 1: This was the only thing I changed in here and it seems like there was some auto-formatting applied.

pinecone/config.py
```diff
@@ -242,18 +243,71 @@ def init(
     project_name: str = None,
     log_level: str = None,
     openapi_config: OpenApiConfiguration = None,
-    config: str = "~/.pinecone",
+    config: str = "./.pinecone",
```
austin-denoble, line 246: I don't think the previous default value worked for having the configuration file in the same location as the repo, but maybe the intent here was different? Tested and this works for me locally.

pinecone/config.py
```diff
@@ -242,18 +242,71 @@ def init(
...
+    The environment variables used to configure the client are the following:
+    
+    ```python
+    export PINECONE_API_KEY="your_api_key"
+    export PINECONE_ENVIRONMENT="your_environment"
+    export PINECONE_PROJECT_NAME="your_project_name"
+    export PINECONE_CONTROLLER_HOST="your_controller_host"
+    ```
+
+    **Using an INI configuration file**
```
austin-denoble, line 270: Slipped a section in here for setting up an INI file. I don't think this was documented at all previously.

PR #220, opened 2023-10-17
Config / init refactor for state encapsulation

jhamon:
## Problem

Singleton config limits how people can use this SDK and makes testing more difficult. Also, we no longer should fetch projectId and construct data url out of configuration parts, to keep the client agnostic to the format of the data plane url.

## Solution

- **State encapsulation using classes**
    - We want to encapsulate all config-related state into a new class `Config` and a subclass `PineconeConfig`.
    - `Config` is a vessel for api_key, host, and openapi configuration. It is host-agnostic because it is used by both the control plane (`Pinecone`) and the data plane (`Index`)
    - `PineconeConfig` is a convenience class to set the controller host for `Pinecone`.
    - New singleton store `IndexHostStore` means we won't make unnecessary and repetitious calls to describe_index if a user creates multiple `Index` instances referring to the same index. In the old paradigm, this was accomplished with a piece of global state storing off projectId.
- **Modified class init for Index** 
    - `Index` class constructor now requires host url instead of an index name. If you use the helper method on a `Pinecone` class instance to instantiate the object, it will manage fetching the host url for you. E.g. `p = Pinecone(api_key='foo'); index = p.Index('my-index')`.

```python
# Old way
import pinecone
pinecone.init(api_key='foo')
index = pinecone.Index('my-index')

# New way
from pinecone import Pinecone
client = Pinecone(api_key='foo')
index = client.Index('my-index')

# New way (alternate, for data-plane only use case)
from pinecone import Index
index = Index(api_key='foo', host='https://data-url')
```

- **Moving things around**
    - `manage.py` got moved and renamed to `pinecone/control/pinecone.py`. The functions such as `create_index`, `describe_index`, etc that previously were top-level package exports relying on global state to access config information have been turned into class methods on a new `Pinecone` class defined in this file.
    - `pinecone/index.py` got moved to `pinecone/data/index.py`. The `pinecone/data` package is a home for all classes used to access the data plane.
    - Everything `grpc` related got shoved into `pinecone/data/grpc`.
    - GRPC mess in one place only. Utils that were grpc-related got pulled out of `pinecone/utils.py` and moved to `pinecone/data/grpc/utils.py`.
    - Leftover util junk moved into a `deprecated` package. Probably will be deleted soon.
- **Breaking down big files into small files**
    - The file previously at `pinecone/index_grpc.py` to split up into many smaller files inside `pinecone/data/grpc/` package. 
    - Remaining util functions in `pinecone/utils.py` got broken up into many smaller files in `pinecone/utils/`
- **Fix existing tests**
    - Needed to adjust a lot of import statements and sometimes setup functions to get pre-existing tests running.  

## Using it

```python
# We're going to use a new Pinecone class instead of interacting directly with the package 
# namespace as in the past
from pinecone import Pinecone

# No more init() method
p = Pinecone(api_key='foo')

# Target an index
index = p.Index('my-index')

# Use data plane methods
index.describe_index_stats()
```

**Mostly working**
- `describe_index` (call succeeds but need to follow up on source_collection field, would ideally return different object for different capacity_modes)

**Somewhat working**:
- `create_index` (API call succeeds, but client blows up when response is not the expected shape)
- `delete_index` (only when timeout=-1, since otherwise it has dependency on `list_indexes`)

**Broken**:
- `list_indexes` (response not expected shape)

## Type of Change

## Test Plan

Automated tests will be expanded in a future PR. A bunch of code moved around to different files or got split from big files into small files and I want to merge this part of the change to unlock parallel work by Austin various follow-ups.


pinecone/__init__.py
```diff
@@ -3,22 +3,10 @@
 """
 from .config import *
 from .exceptions import *
-from .manage import *
-from .index import *
+from .control.pinecone import Pinecone
+from .data.index import *
 
 try:
```
jhamon, line 9: These `try` import statements led to some pretty confusing debug situations when stuff started failing due to me moving packages from one location to another since the intention is really only to ignore failures caused by people installing without grpc extras. We should revisit this pattern and make sure we're testing stuff both with and without grpc deps loaded.

pinecone/__init__.py
```diff
@@ -3,22 +3,10 @@
...
-    from .grpc.index_grpc import *
+    from .data.grpc.index_grpc import *
 except ImportError:
     pass  # ignore for non-[grpc] installations
```
jhamon, line 12: These `try` import statements led to some pretty confusing debug situations when stuff started failing due to me moving packages from one location to another since the intention is really only to ignore failures caused by people installing without grpc extras. We should revisit this pattern and make sure we're testing stuff both with and without grpc deps loaded.

pinecone/config/logging.py
jhamon, line 1: Moved these constants out of old `utils.py` since they were only being used in one place.

The rest of this I pulled out of We need a whole ticket to revisit and overhaul the logging approach in a holistic way.

pinecone/config/openapi.py
jhamon, line 1: Everything in here was pretty much copy/pasta out of the old config.py. I haven't done anything to verify these settings for openapi, but they worked in the past so hopefully they are sufficient for now.

pinecone/config/pinecone_config.py
```diff
@@ -0,0 +1,10 @@
+import os
+from .config import Config
+
+DEFAULT_CONTROLLER_HOST = "https://api.pinecone.io"
+
+
+class PineconeConfig(Config):
+    def __init__(self, api_key: str = None, host: str = None, **kwargs):
+        host = host or kwargs.get("host") or os.getenv("PINECONE_CONTROLLER_HOST") or DEFAULT_CONTROLLER_HOST
```
jhamon, line 9: I don't know if we want to keep this or not, but in the past there was an env var called `PINECONE_CONTROLLER_HOST` being used.

But as you can see, the only difference between a `Config` instance and the `PineconeConfig` subclass is a little bit of logic here about how to set the host value.

pinecone/control/index_host_store.py
```diff
@@ -0,0 +1,39 @@
...
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            instance = super().__call__(*args, **kwargs)
+            cls._instances[cls] = instance
+        return cls._instances[cls]
+
+
+class IndexHostStore(metaclass=SingletonMeta):
```
jhamon, line 15: I confess, I had help from ChatGPT on this. I still need to add tests to verify the singleton is working as expected.

pinecone/control/models/index_description.py
```diff
@@ -0,0 +1,13 @@
+from typing import NamedTuple
```
jhamon, line 1: Pulled this out of legacy `manage.py`. Not sure if we really want to keep it or not long term.

pinecone/control/pinecone.py
jhamon, line 1: Most of the logic in this file moved from `manage.py`

pinecone/control/pinecone.py
```diff
@@ -0,0 +1,277 @@
...
+
+        :param name: the name of the index to describe.
+        :return: Returns an `IndexDescription` object
+        """
+        api_instance = self.index_api
+        response = api_instance.describe_index(name)
+        db = response.database
+        host = response.status.host
+
+        self.index_host_store.set_host(self.config, name, "https://" + host)
```
jhamon, line 196: We proactively stick this host url in the singleton store if they happen to call describe_index before doing any index operations. Just to save an additional round trip.

pinecone/core/client/api/index_operations_api.py
```diff
@@ -134,7 +134,7 @@ def configure_index_with_http_info(
         :rtype: tuple(IndexMeta, status_code(int), headers(HTTPHeaderDict))
         """
 
-        _hosts = ["https://controller.{environment}.pinecone.io"]
+        _hosts = ["https://api.pinecone.io"]
```
jhamon, line 137: Changes in this file are generated, derived from changes to the `servers` field in our openapi spec stored in the other repository.

pinecone/core/client/models/index_meta_database.py
```diff
@@ -30,7 +30,7 @@ class IndexMetaDatabase(BaseModel):
     """
 
     name: StrictStr = Field(...)
-    dimension: StrictStr = Field(...)
+    dimension: StrictInt = Field(...)
```
jhamon, line 33: This is also a generated change, based on an updated openapi spec.

pinecone/data/grpc/base.py
```diff
@@ -0,0 +1,192 @@
+import logging
```
jhamon, line 1: Moved these out of the giant `grpc_index.py` file. No changes.

pinecone/data/grpc/config.py
```diff
@@ -0,0 +1,33 @@
+from .retry import RetryConfig
```
jhamon, line 1: Moved these out of the giant `grpc_index.py` file. No changes.

pinecone/data/grpc/future.py
```diff
@@ -0,0 +1,34 @@
+from grpc._channel import _MultiThreadedRendezvous
```
jhamon, line 1: Moved these out of the giant `grpc_index.py` file. No changes.

pinecone/data/grpc/index_grpc.py
```diff
@@ -1,22 +1,19 @@
 import logging
 import numbers
-from abc import ABC, abstractmethod
-from functools import wraps
-from importlib.util import find_spec
-from typing import NamedTuple, Optional, Dict, Iterable, Union, List, Tuple, Any
+from typing import Optional, Dict, Iterable, Union, List, Tuple, Any
```
jhamon, line 3: Extracted a bunch of supporting classes and functions into smaller files. Removed dependencies no longer in use by the class defined in this file. Otherwise, no changes.

pinecone/data/grpc/utils.py
```diff
@@ -0,0 +1,107 @@
+import uuid
```
jhamon, line 1: Moved these grpc-related utils out of a general `utils.py` file so there wouldn't be any try/except import statements needed in here, and also so they are closer to where they are actually being used in the code.

tests/unit/test_manage.py
jhamon, line 1: This file got renamed and modified. Now it is `tests/unit/test_control.py`

pinecone/utils/user_agent.py
```diff
@@ -0,0 +1,10 @@
+import requests
```
jhamon, line 1: Moved from `pinecone/utils.py`

pinecone/utils/fix_tuple_length.py
jhamon, line 1: Moved from `pinecone/utils.py`

pinecone/utils/error_handling.py
jhamon, line 1: Moved from `pinecone/utils.py`

pinecone/utils/environment.py
jhamon, line 1: Moved from `pinecone/utils.py`

pinecone/utils/deprecation_notice.py
jhamon, line 1: Moved from `pinecone/utils.py`

pinecone/utils/constants.py
jhamon, line 1: Moved some constants closer to where they are actually being used, since they were only used in one spot.

pinecone/utils/check_kwargs.py
jhamon, line 1: Moved from `pinecone/utils.py`

pinecone/deprecated/legacy_utils.py
```diff
@@ -0,0 +1,88 @@
+import re
```
jhamon, line 1: Leftovers, hopefully to be deleted soon pending investigation on vector_column_service.proto

pinecone/data/index.py
```diff
@@ -81,25 +82,30 @@ def upsert_numpy_deprecation_notice(context):
...
+        api_client = ApiClient(configuration=self._config.OPENAPI_CONFIG, pool_threads=pool_threads)
+        api_client.user_agent = get_user_agent()
+        self._api_client = api_client
+        self._vector_api = VectorOperationsApi(api_client=api_client)
+    
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self._api_client.close()
```
jhamon, line 109: These methods were the only thing I needed to add back as a consequence changing `Index` to not subclass `ApiClient` to satisfy one test (using `with Index(...) as index` syntax). The [python language docs](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement) show tthese must be implemented in order to use the `with` syntax.

pinecone/data/index.py
```diff
@@ -81,25 +82,30 @@ def upsert_numpy_deprecation_notice(context):
...
+    def __init__(self, api_key: str, host: str, pool_threads=1, **kwargs):
+        api_key = api_key or kwargs.get("api_key")
+        host = host or kwargs.get('host')
+        pool_threads = pool_threads or kwargs.get("pool_threads")
+
+        self._config = Config(api_key=api_key, host=host, **kwargs)
+        
+        api_client = ApiClient(configuration=self._config.OPENAPI_CONFIG, pool_threads=pool_threads)
+        api_client.user_agent = get_user_agent()
+        self._api_client = api_client
```
jhamon, line 101: This is a substantive change to adopt the new `Config` class.

pinecone/utils/error_handling.py
```diff
@@ -3,19 +3,14 @@
 
 from urllib3.exceptions import MaxRetryError, ProtocolError
 
-from pinecone import Config
-
 def validate_and_convert_errors(func):
     @wraps(func)
     def inner_func(*args, **kwargs):
-        Config.validate()  # raises exceptions in case of invalid config
```
austin-denoble, line 11: This happens within the init of the `Config` itself, right? Not really necessary to have it here.

tests/unit/test_config.py
```diff
@@ -1,209 +1,80 @@
...
-    controller_host = "kwargs-controller-host"
-
-    pinecone.init(environment=env, api_key=api_key, host=controller_host)
-
-    assert Config.API_KEY == api_key
-    assert Config.ENVIRONMENT == env
-    assert Config.CONTROLLER_HOST == controller_host
-
-
-def test_resolution_order_kwargs_over_config_file():
```
austin-denoble, line 135: It looks like we lost the tests around resolution order over a config file, but I'd assume that's fine given we'll need to add new testing around the config singleton and stuff anyways, and the config file setup isn't really documented.

austin-denoble (approved):
Thank you so much for taking this on! Organizationally and architecturally this feels much, much better to me. I think the refactor of the singleton approach and the ability for users to be more flexible when using the client is a big win.

Overall this looks good to go I think. I'd like to do some detailed testing around the changes here but will not have time this evening.

Definitely would like to get this merged to the feature branch and start iterating on it.

pinecone/control/models/collection_description.py
```diff
@@ -0,0 +1,7 @@
+class CollectionDescription(object):
```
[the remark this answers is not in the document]
jhamon: Whether it's a dictionary or an extention of `NamedTuple`, the usual motivation for a presenter object like is just to have greater control over what is shown to the user and discard any unwanted fields that may be emitted by the underlying generated code. I don't see why that's needed in this case though. Probably I will delete it soon but not in this diff.

PR #218, opened 2023-10-12
Support passing custom host for Index operations

austin-denoble:
## Problem
We need support for supplying a custom `host` value for index operations.

## Solution
`pinecone.init()` and `Config` already support the ability to supply a custom `host`. It didn't seem like it was being passed through to the `ApiClient`:

- Update `_get_api_instance()` in `manage.py` to check for a `Config.CONTROLLER_HOST` value, and apply it to the `client_config` if so. This will then be used throughout the index operations.
- Add a unit test file for `manage.py`, test setting `controller_host` and timeout functionality in `create_index` / `delete_index`.

## Type of Change
- [X] New feature (non-breaking change which adds functionality)

## Test Plan
Pull this branch down and test locally. I validated myself by passing a bogus `host` in and looking over the errors, then passing the proper `host` via `init` to validate it works properly.

```python
>>> poetry shell
>>> python3

>>> import pinecone
>>> pinecone.init(api_key="123-456-789", environment="my_environment", host: "https:custom-host-foo.com")

# Having passed a custom host that's inaccessible should fail
>>> pinecone.list_indexes()
# urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='fooblahblah.io', port=443): Max retries exceeded with  
# url: /databases (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x107eb9dd0>:
# Failed to resolve 'fooblahblah.io' ([Errno 8] nodename nor servname provided, or not known)"))


# Re-Init - should work as expected
>>> pinecone.init(api_key="123-456-789", environment="my_environment")

pinecone.list_indexes()
# ['austin-3', 'austin-py-test', 'resin--test-index-1', 'test-create-1']
```

PR #221, opened 2023-10-17
Prevent _get_api_instance from modifying OpenAPI config

austin-denoble:
## Problem
Previous revision we added some code to apply a custom `host` value to control plane operations: https://github.com/pinecone-io/pinecone-python-client/pull/218

This change caused 404s when calling dataplane operations. The `Configuration._base_path` value was being overwritten by the custom `host` that was provided due to the `Configuration.host` setter:

```
  @host.setter
  def host(self, value):
      """Fix base path."""
      self._base_path = value
      self.server_index = None
```
https://github.com/pinecone-io/pinecone-python-client/blob/7ff4ed877ac3e97eb557b747a072a84e57764b91/pinecone/core/client/configuration.py#L493

## Solution
The change I made was actually modifying the reference to `Config.OPENAPI_CONFIG` which overwrote things across the board.

Instead of operating on the singleton's instance of the config, we can use `copy.deepcopy(Config.OPENAPI_CONFIG)` and use that to construct things before handing it off to `ApiClient`

## Type of Change

## Test Plan
Pull this branch down, use the client locally and make sure that dataplane calls work properly without a `host` or with one in the case of supplying a custom control plane URL.

### DIFF
--- a/pinecone/manage.py
+++ b/pinecone/manage.py
@@ -1,5 +1,6 @@
 import time
 from typing import NamedTuple, Optional
+import copy
 
 import pinecone
 from pinecone.config import Config
@@ -49,7 +50,7 @@ class CollectionDescription(object):
 
 
 def _get_api_instance():
-    client_config = Config.OPENAPI_CONFIG
+    client_config = copy.deepcopy(Config.OPENAPI_CONFIG)
     client_config.api_key = client_config.api_key or {}
     client_config.api_key["ApiKeyAuth"] = client_config.api_key.get("ApiKeyAuth", Config.API_KEY)
     client_config.server_variables = {**{"environment": Config.ENVIRONMENT}, **client_config.server_variables}


pinecone/manage.py
```diff
@@ -49,7 +50,7 @@ def __str__(self):
 
 
 def _get_api_instance():
-    client_config = Config.OPENAPI_CONFIG
+    client_config = copy.deepcopy(Config.OPENAPI_CONFIG)
```
austin-denoble, line 53: This is what we're doing in `index.py`, I think it's probably prudent to do that here as well rather than possibly changing fields on the singleton.

https://github.com/pinecone-io/pinecone-python-client/blob/7ff4ed877ac3e97eb557b747a072a84e57764b91/pinecone/index.py#L92

jhamon (approved):

PR #223, opened 2023-10-19
Add control plane integration tests

austin-denoble:
## Problem
Our Python client doesn't have true integration tests. Currently, we have tests under `/integ/test_index.py` which mock all the requests and responses and primarily stress the connections between various elements of the client codebase. Fleshing out integration tests would help us have more confidence when making changes to the Python client, especially when involving the network layer.

## Solution
The plan is to split between control plane and data plane tests. For PRs I thought it'd be reasonable to split things out, and then finish off with CI changes to bring everything together.

- Add a new section for integration tests under `/tests/integration/control/`
- Implement control plane integration tests:
    - `test_create_index.py`
    - `test_configure_index.py`
    - `test_describe_index.py`
    - `test_list_indexes.py`

Currently, the entire control plane suite completes in about ~35 seconds or so. `sdk-python-testing` currently has a Pod Limit of 7, and these seem to complete fine, but I'd like to get quote bumped to reduce possibility of flaking.

Screenshot 2023-10-19 at 1 45 55 PM

**Question: ** Should I look at rolling some collection tests here as well?

## Type of Change

## Test Plan
Currently, these tests aren't being run in CI. That will come in a following PR. The environment variables that are used will be setup in GitHub secrets for the repo/workflows.

To test locally, you'll need to set environment variables and then call `pytest`:

```
$ export PINECONE_API_KEY="my-api-key"
$ export PINECONE_ENVIRONMENT="my-environment"

$ poetry run pytest tests/integration/control/
...
```

### DIFF
--- a/tests/integration/control/test_configure_index.py
+++ b/tests/integration/control/test_configure_index.py
@@ -1,0 +1,76 @@
+import pytest
+import pinecone
+import os
+import tests.test_helpers as test_helpers
+
+
+class TestConfigureIndex:
+    @pytest.fixture(autouse=True)
+    def setup_and_teardown(self):
+        # setup
+        api_key = os.getenv("PINECONE_API_KEY")
+        environment = os.getenv("PINECONE_ENVIRONMENT")
+
+        self.index_name = test_helpers.generate_index_name()
+        self.pinecone = pinecone
+        self.pinecone.init(api_key=api_key, environment=environment)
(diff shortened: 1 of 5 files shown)


jackpertschuk:
is this for new version of the SDKs or old?

austin-denoble:
> is this for new version of the SDKs or old?

Old technically, but we'll be updating them around the changes for the new SDK.

PR #229, opened 2023-10-26
Downgrade to openapi 5.2, add template overrides

jhamon:
## Problem

Recent versions of openapi generator have added dependencies we don't want to take on, especially `pydantic` v1. Pydantic v2 has been out for a while and we think this is very likely to lead to dependency clashes for some users.

## Solution

The reason we updated in the first place was to avoid generating code that emitted urllib3 deprecation notices. In this PR, I downgrade to the previously used 5.2.0 generator version and address the urllib3 error using [template overrides](https://openapi-generator.tech/docs/templating/) in the generation step.

In this diff:
- Go back to generating the core with openapi 5.2.0 (same version as before, e.g. on main)
  - Everything under `pinecone/core` is generated using openapi and don't need to be specifically reviewed. This is like 90% of the files changed.
- Use a template override in the generation step (handled in a private repo) to adjust the urllib3 getheaders() usage and avoid spamming those deprecation notices.
    - Notice that generated `rest.py` is using [correct syntax](https://stackoverflow.com/questions/75382508/deprecationwarning-httpresponse-getheader-is-deprecated-and-will-be-removed-i) for getting headers in urllib3 
- Fix tests that broke along the way, mostly due to changes in package structure used by the generated code.
- Remove dependencies from `pyproject.toml` that were added when previously going from openapi 5.2.0 => 7.0.1 (pydantic, aenum)

## Type of Change

- [x] None of the above: Update generated code, change dependencies

## Follow-up needed

- Docs build broken. By the looks of it, it seems related to numpy. But I want to remove numpy in a follow-up diff anyway so I may merge with this build failing for now.


pinecone/core/client/rest.py
```diff
@@ -1,59 +1,49 @@
...
         self.data = resp.data
 
     def getheaders(self):
         """Returns a dictionary of the response headers."""
-        return self.urllib3_response.headers
+        return self.urllib3_response.getheaders()
 
     def getheader(self, name, default=None):
         """Returns a given response header."""
         return self.urllib3_response.headers.get(name, default)
```
jhamon, line 41: This `return self.urllib3_response.headers.get(name, default)` is not showing as a change because it was handled correctly in the 7.0.1 generator that we are downgrading from. But the fact it is not changed means my template override was successfully used; without this template override, openapi 5.2.8 generates `return self.urllib3_response.getheader(name, default)` which causes a large amount of runtime deprecation notices to appear.

pinecone/data/index.py
```diff
@@ -69,10 +69,10 @@
...
 def parse_query_response(response: QueryResponse, unary_query: bool):
     if unary_query:
-        response.results = None
+        response._data_store.pop("results", None)
     else:
-        response.matches = None
-        response.namespace = None
+        response._data_store.pop("matches", None)
+        response._data_store.pop("namespace", None)
     return response
```
jhamon, line 76: I don't like this at all, but I brought it back from [`main`](https://github.com/pinecone-io/pinecone-python-client/blob/main/pinecone/index.py#L69) to get some tests passing. It seems this thing is supposed to return a different data shape depending on what query parameters were used, which is a terrible design. But not something I want to address right now in this diff with all the generated changes.

tests/unit/test_index.py
```diff
@@ -3,7 +3,6 @@
 import pytest
 
 import pinecone
-from pinecone.config import PineconeConfig
```
jhamon, line 6: I noticed this wasn't being used, so I deleted it.

tests/unit/test_control.py
```diff
@@ -24,43 +23,24 @@ def test_passing_host(self):
         (-1, [{"status": {"ready": False}}], 0, 0),
     ])
     def test_create_index_with_timeout(self, mocker, timeout_value, describe_index_responses, expected_describe_index_calls, expected_sleep_calls):
-        mocker.patch.object(IndexOperationsApi, 'describe_index', side_effect=describe_index_responses)
```
jhamon, line 27: The IndexOperationsApi class generated by openapi 5.2.8 can't be mocked in the normal way because almost all of its methods are attributes defined at runtime from within __init__(). I'm not a mocker guru by any means, but the only way I could get this working was to grab the instance of IndexOperationsApi being used by the pinecone client and monkey patch that directly. 🤦

tests/unit/test_control.py
```diff
@@ -24,43 +23,24 @@ def test_passing_host(self):
...
         with pytest.raises(TimeoutError):
-            get_status_responses = [{"status": {"ready": False}}] * 5
-            mocker.patch.object(IndexOperationsApi, 'describe_index', side_effect=get_status_responses)
-            mocker.patch.object(IndexOperationsApi, 'create_index')
-            mocker.patch('time.sleep')
-
             p = Pinecone(api_key="123-456-789")
-            p.create_index("my-index", 10, timeout=10, cloud="aws", region="us-west1", capacity_mode="pod")
-
-    # @pytest.mark.parametrize("timeout_value, list_indexes_calls, time_sleep_calls, list_indexes_responses", [
```
jhamon, line 48: This commented code was duplicating a different test, so i just removed it.

austin-denoble (approved):
Looks good, thanks a ton for taking this on to get things back in place to avoid the dependency hell problem. :shipit:

austin-denoble:
The docs build failure is confusing to me since we're also running `poetry install` for the testing steps and that seems to complete fine, and I don't think they use differing dependencies.

pinecone/config/openapi.py
```diff
@@ -17,7 +17,7 @@
 class OpenApiConfigFactory:
     @classmethod
     def build(cls, api_key: str, host: str = None, **kwargs):
-        openapi_config = OpenApiConfiguration.get_default()
```
rohanshah18, line 20: What's the reason for this change?

rohanshah18 (approved):
LGTM, thanks for making this change!

pinecone/config/openapi.py
jhamon: This method doesn't exist in the generated output of openapi 5.2. It shouldn't matter, since we're setting up all the important properties within this config factory class anyway.

tests/unit/test_control.py
jhamon: I thought it was duping `test_create_index_with_timeout` (similar parameterized setup, the word "timeout"), but actually now that I am looking closer I see it doing delete_index instead of create_index. I should add more tests here, but I can do that in a separate follow-up.

Issue #148, opened 2023-03-12
Deprecate Numpy


philipk19238:
I'm requesting to remove numpy as a dependency. It is barely used within the repository and creates such a large dependency that is unexpected from what is essentially an API wrapper. Anyone who is trying to deploy a lambda function to AWS that uses the pinecone client would have to create a custom numpy layer to support the pinecone client.

tsah-alike:
We are using pinecone client from AWS lambda and we are unable to upgrade because of this issue. I suggest adding support in a dedicated package (as with grpc)

daverigby:
This is fixed by #233 which moves Numpy to be a development-only dependancy (it's used in various unit tests). This was included in [v3.0.0](https://github.com/pinecone-io/pinecone-python-client/releases/tag/v3.0.0).

PR #233, opened 2023-11-02
Remove numpy dependency

jhamon:
## Problem

Our numpy dependency causes installation problems for some users and we don't really need it.

## Solution

- Dependency changes
    - Move `numpy` from `dependencies` into `dev.dependencies` in `pyproject.toml`.
    - We still need it as a dev dependency so we can use it in tests to make sure things don't blow up when users pass in numpy arrays to the `upsert` method.
- Coercing inputs to `list` from pandas, numpy arrays
    - Implement a util function, `convert_to_list` that will cast iterables from numpy, pandas, etc to a list without any specific dependency on these packages.
- Parsing user inputs to upsert
    - Extract a big chunk of code from `Index#upsert` and move it into a new `VectorFactory` class where it could be tested in isolation. This seemed warranted because there were many different error cases being covered in this code that parses user inputs into a `Vector` ahead of the upsert API call. (This became part of this PR because buried in the guts of it it were some numpy usages.)
    - Add lots of missing tests for `VectorFactory` code.
- Take out the trash
    - Delete some legacy utils and references messing with numpy and `vector_column_service.proto` because these seemed completely disconnected from the rest of the Index/GRPCIndex code.
    - Remove numpy deprecation notices

## Type of Change

- [x] Breaking change: Removing old utils functions that seem useless anyway
- [x] None of the above: Refactor


pinecone/data/index.py
```diff
@@ -168,7 +154,7 @@ def upsert(
 
         Returns: UpsertResponse, includes the number of vectors upserted.
         """
-        _check_type = kwargs.pop("_check_type", False)
```
jhamon, line 171: I suppose this is a breaking change also. Type checking seems like something we want to happen by default.

pinecone/data/index.py
```diff
@@ -197,81 +183,11 @@ def _upsert_batch(
...
-                        f"Found a tuple of length {len(item)} which is not supported. "
-                        f"Vectors can be represented as tuples either the form (id, values, metadata) or (id, values). "
-                        f"To pass sparse values please use either dicts or a Vector objects as inputs."
-                    )
-                id, values, metadata = fix_tuple_length(item, 3)
-                return Vector(id=id, values=values, metadata=metadata or {}, _check_type=_check_type)
-            elif isinstance(item, Mapping):
-                return _dict_to_vector(item)
-            raise ValueError(f"Invalid vector value passed: cannot interpret type {type(item)}")
+        vec_builder = lambda v: VectorFactory.build(v, check_type=_check_type)
```
jhamon, line 186: `VectorFactory.build` is static, but we need to wrap it in this lambda to thread through the `check_type` param when mapping over all the vectors passed to upsert.

pinecone/data/vector_factory.py
```diff
@@ -0,0 +1,109 @@
+import numbers
```
jhamon, line 1: This began as a copy/paste of the removed code in `Index` but I refactored it a good bit to cover some cases that weren't being handled.

pinecone/utils/convert_to_list.py
```diff
@@ -0,0 +1,9 @@
+def convert_to_list(obj):
+    class_name = obj.__class__.__name__
+
+    if class_name == 'list':
+        return obj
+    elif hasattr(obj, 'tolist'):
+        return obj.tolist()
+    else:
+        return list(obj)
```
jhamon, line 9: All classes that implement `__iter__` methods can be passed to the builtin python `list()` method, but calling `list()` on a numpy ndarray gives back a list of `int64` objects which is numpy-specific and causes the type checking in `Vector` to blow up. 

Numpy's `Ndarray` and pandas' `IntegerArray`, `Serial`, etc have an alternative method called `tolist()` that returns a list of native python `int`s instead. So in this function we just check if that method is present and call it if it is. This lets us to the conversion without creating an explicit dependency like before in places where we were checking if something was a numpy object with `isinstance`.

pinecone/data/index.py
```diff
@@ -197,81 +183,11 @@ def _upsert_batch(
...
-
-                indices = item["sparse_values"].get("indices", None)
-                values = item["sparse_values"].get("values", None)
-
-                if isinstance(values, np.ndarray):
-                    upsert_numpy_deprecation_notice("Deprecated type passed in sparse_values['values'].")
-                    values = values.tolist()
-                if isinstance(indices, np.ndarray):
-                    upsert_numpy_deprecation_notice("Deprecated type passed in sparse_values['indices'].")
-                    indices = indices.tolist()
```
jhamon, line 229: This numpy stuff buried here in the guts is why I did the class extraction and added a bunch of tests for `VectorFactory`. Relying on `isinstance` like this created a dependency on numpy that wasn't really needed to accomplish the same result.

tests/unit/data/test_vector_factory.py
```diff
@@ -0,0 +1,115 @@
...
+        d = {'id': '1', 'values': [0.1, 0.2, 0.3], 'metadata': {'genre': 'comedy'}, 'sparse_values': {'indices': [0, 2], 'values': np.array([0.1, 0.3])}}
+        actual = VectorFactory.build(d)
+        expected = Vector(id='1', values=[0.1, 0.2, 0.3], metadata={'genre': 'comedy'}, sparse_values=SparseValues(indices=[0, 2], values=[0.1, 0.3]))
+        assert actual == expected
+
+    def test_build_when_dict_sparse_values_when_indices_is_ndarray(self):
+        d = {'id': '1', 'values': [0.1, 0.2, 0.3], 'metadata': {'genre': 'comedy'}, 'sparse_values': {'indices': np.array([0, 2]), 'values': [0.1, 0.3]}}
+        actual = VectorFactory.build(d)
+        expected = Vector(id='1', values=[0.1, 0.2, 0.3], metadata={'genre': 'comedy'}, sparse_values=SparseValues(indices=[0, 2], values=[0.1, 0.3]))
+        assert actual == expected
```
austin-denoble, line 111: Can users also pass a pandas array for `indices` / `values` and should that be tested as well?

tests/unit/test_index.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 
 import pinecone
-from pinecone import Index
+from pinecone import Index, ApiTypeError
```
austin-denoble, line 6: Is `ApiTypeError` needed here?

pinecone/data/vector_factory.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self, item):
+        message = f"Column `metadata` is expected to be a dictionary, found {type(item['metadata'])}"
+        super().__init__(message)
+
+class VectorFactory:
+    @staticmethod
+    def build(item: Union[Vector, Tuple, Dict], check_type: bool = True) -> Vector:
+        if isinstance(item, Vector):
+            return item
+        elif isinstance(item, tuple):
```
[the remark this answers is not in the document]
jhamon: You'd think so, but the class name is lowercase.

```
>>> x = (1,2)
>>> x.__class__
<class 'tuple'>
```

tests/unit/data/test_vector_factory.py
jhamon: Yeah, they can since `convert_to_list` handles both. I'll add a few more tests.

pinecone/data/vector_factory.py
```diff
@@ -0,0 +1,109 @@
...
+        if "metadata" in item:
+            metadata = item.get("metadata")
+            if not isinstance(metadata, Mapping):
+                raise MetadataDictionaryExpectedError(item)
+
+        try:
+            return Vector(**item, _check_type=check_type)
+        except TypeError as e:
+            # if not isinstance(item['values'], Iterable) or not isinstance(item['values'][0], numbers.Real):
+            #     raise TypeError(f"Column `values` is expected to be a list of floats")
```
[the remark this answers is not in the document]
jhamon: Yeah, copy/pasta leftovers. I'll delete.

daverigby:
Fixes #148.

PR #234, opened 2023-11-02
Pin docs-build python version

jhamon:
## Problem

The build-docs workflow action has been failing since `setup-python`, which was configured with floating version `3.x`, rolled over to install the new 3.12 cpython release.

## Solution

Pin to cpython 3.11 so things can build while we do more investigation about how to unblock running with 3.12.

## Type of Change

## Test Plan

See builds passing again.


austin-denoble (approved):
🚢

PR #240, opened 2023-11-04
Add static type checking with mypy

jhamon:
## Problem

Some parts of this code were including typehints, but we haven't been running any checks to verify the correctness of those annotations. Static typechecking helps to improve overall quality and catch mistakes.

## Solution

- Add a dev dependency on `mypy` ([site](https://mypy-lang.org/)) and packages with types for third-party dependencies.
- Make adjustments throughout the code to satify the typechecker
- Add steps in CI to run mypy checks automatically on every change.
- For now, exclude generated code in `pinecone/core` because it has a bunch of issues but is not easy to change.

Summary of code changes:
- Everywhere that has a default value of `None` needs to be wrapped in the `Optional` type from the `typings` package. Correct: `def my_method(foo: Optional[str] = None)`. Incorrect: `def my_method(foo: str = None)`
- Refactored config-related stuff a bit because having config classes with inner `_config` attributes seemed like a smell and also made typing config-related stuff more challening. Ended up converting what was previously the `Config` class into a `ConfigBuilder` and the actual config object is now just a `NamedTuple` which is a container for immutable data.
- Mypy doesn't like when you overwrite variables in a way that changes the type. So some places required me to introduce an additional intermediate variable to satisfy the type checker.
- Mypy got mad about having a `Pinecone` class in both `pinecone` and `pinecone.grpc`. So I renamed the GRPC one to `PineconeGRPC`.

## Type of Change

## Test Plan

See new step running in CI.


.github/workflows/testing.yaml
```diff
@@ -24,6 +24,13 @@ jobs:
 
       - name: Run unit tests
         run: poetry run pytest --cov=pinecone --timeout=120 tests/unit
+      
+      - name: mypy check
+        run: |
+          # Still lots of errors when running on the whole package (especially
+          # in the generated core module), but we can check these subpackages
+          # so we don't add new regressions.
+          poetry run mypy pinecone --exclude pinecone/core --exclude pinecone/grpc
```
rohanshah18, line 33: Is pinecone/grpc also auto-generated and the reason for exclusion?

pinecone/data/index.py
```diff
@@ -180,7 +180,7 @@ def upsert(
         return UpsertResponse(upserted_count=total_upserted)
 
     def _upsert_batch(
-        self, vectors: List[Vector], namespace: Optional[str], _check_type: bool, **kwargs
+        self, vectors: Union[List[Vector], List[tuple], List[dict]], namespace: Optional[str], _check_type: bool, **kwargs
```
rohanshah18, line 183: What's the reason for this change here?

pinecone/data/vector_factory.py
```diff
@@ -91,7 +91,7 @@ def _dict_to_vector(item, check_type: bool) -> Vector:
         try:
             return Vector(**item, _check_type=check_type)
         except TypeError as e:
-            if not isinstance(item["values"], Iterable) or not isinstance(item["values"][0], numbers.Real):
+            if not isinstance(item["values"], Iterable) or not isinstance(item["values"].__iter__().__next__(), numbers.Real):
```
rohanshah18, line 94: Isn't item["values"][0] better in terms of readability? What's the reason for this change?

pyproject.toml
```diff
@@ -54,11 +54,11 @@ classifiers=[
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.8,<4.0"
+python = ">=3.9,<3.13"
```
rohanshah18, line 57: Whats the reason for this upper bound set to <3.13?

pinecone/data/index.py
jhamon: The `upsert()` method that calls this `_upsert_batch()` takes a wide variety of user inputs (`Vector`, `tuple`, or `dict` representations of vectors) which aren't normalized to`Vector` until passed through the `VectorBuilder` referenced a few lines into this method. So `List[Vector]` is not a correct description of the input types accepted.

rohanshah18: Ok got it thanks!

rohanshah18 (approved):
Overall the changes looks good to me. I have left a few questions where I wasn't clear about the change.

pinecone/data/vector_factory.py
jhamon: It was better readability, but it was a type error because in python something that is `Iterable` is not necessarily subscriptable with `[0]`. 

To be `Iterable` something must implement the `__iter__()` method and returns an iterator (i.e. something implementing `__next__()`). In normal usage, these methods are not referenced directly... much more common to see iterators used in list comprehensions or in for-loop statements.

To use indexes like `foo[0]`, a class must implement the magic method `__getitem__()`

So this was a type error because not all iterables support indexing like `foo[0]`. Even though python lists (basically arrays) are both iterable and indexable, that's not always the case for all `Iterable` items from the compiler's perspective.

Relevant docs: 
- https://wiki.python.org/moin/Iterator
- https://docs.python.org/3/reference/datamodel.html#object.__getitem__

austin-denoble: Thanks for the explanation. 👍

pyproject.toml
jhamon: 3.13 is the highest python version that currently exists. Poetry complained that some of the dependencies added (it was types for pandas, I think) were not compatible with python versions greater than 3.13. So I adjusted the range to reflect the current reality.

pyproject.toml
```diff
@@ -77,15 +77,15 @@ pytest-timeout = "1.4.2"
...
+types-protobuf = "^4.24.0.4"
 
 [tool.poetry.extras]
-# Poetry's concept of extras is what allows us to separate the gRPC dependencies, 
-# which will only be installed if you run `poetry install --extras "grpc"`,
-# from the base dependencies. 
-#
-# Note that Poetry expects the dependencies to be defined in either tool.poetry.dependencies 
-# or tool.poetry.group.dev.dependencies, but they're only referenced by name in the grpc entry under 
-# tool.poetry.extras
```
austin-denoble, line 88: Thanks for pulling this out.

austin-denoble (approved):
Looks good to me, this will be a big boon for working in this repo. Thanks for the detailed answers to @rohanshah18's questions. 🚢

PR #245, opened 2023-11-18
[spruce] Add additional_headers config

jhamon:
## Problem

To support API development and testing, we would like a way to pass additional headers with every request.

## Solution

Add a new `additional_headers` config option and wire up the api client internals to use them.

Usage:

```python
from pinecone import Pinecone

p = Pinecone(api_key='asdf', additional_headers={'X-SECRET-EXTRAS': 'secret secret'})
```

## Type of Change

## Test Plan

Added a unit test and also verified manually running with `PINECONE_DEBUG_CURL=true`

Screenshot 2023-11-18 at 5 53 56 PM

### DIFF
--- a/pinecone/config/config.py
+++ b/pinecone/config/config.py
@@ -1,4 +1,4 @@
-from typing import NamedTuple, Optional
+from typing import NamedTuple, Optional, Dict
 import os
 
 from pinecone.exceptions import PineconeConfigurationError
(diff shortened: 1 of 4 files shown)


austin-denoble (approved):
Looks good. 🚢 

I'm assuming this is primarily useful in the context of the Python client being the client of choice for Platform, but if we'd like something similar on the TypeScript side I can take a look.

PR #246, opened 2023-11-21
[Fix] Prelease CI job uses bump action to version

jhamon:
## Problem

When we migrated to `poetry`, the prerelease job did not get updated to set prerelease versions correctly.

## Solution

We need to use our action to compute the correct prerelease version.

## Type of Change


austin-denoble (approved):
🚢

PR #247, opened 2023-11-22
[spruce] Make list results iterable

jhamon:
## Problem

Results from list_indexes a bit annoying to work with because the object isn't iterable

## Solution

Proxy the object and add methods to make it iterable. This gives most of the benefits of drilling into the response and returning the list under the `indexes` while maintaining the flexility to access any attributes that get added to `IndexList` in the future (e.g. for pagination)

## Usage

```python
>>> p = Pinecone(api_key='API_KEY')
>>> results = p.list_indexes()
>>> results.__class__
<class 'pinecone.models.iterable_index_list.IterableIndexList'>
```

Use this object as if it a normal list
```
>>> len(results)
2
>>> results[0]
{'dimension': 3,
 'host': 'jen3-sl-126ib2q.svc.dev-aws-ram-paid.pinecone.io',
 'metric': 'cosine',
 'name': 'jen3-sl',
 'spec': {'serverless': {'cloud': 'aws', 'region': 'us-west-2'}},
 'status': {'ready': True, 'state': 'Ready'}}
>>> names = [x.name for x in results]
>>> names
['jen3-sl', 'jen2-sl']
```

However the representation in the console reflects that it is an object with attributes. And those attributes can still be accessed (e.g. `results.indexes`)
```
>>> results.indexes
[{'dimension': 3,
 'host': 'jen3-sl-126ib2q.svc.dev-aws-ram-paid.pinecone.io',
 'metric': 'cosine',
 'name': 'jen3-sl',
 'spec': {'serverless': {'cloud': 'aws', 'region': 'us-west-2'}},
 'status': {'ready': True, 'state': 'Ready'}}, {'dimension': 2,
 'host': 'jen2-sl-126ib2q.svc.apw5-4e34-81fa.pinecone.io',
 'metric': 'cosine',
 'name': 'jen2-sl',
 'spec': {'serverless': {'cloud': 'aws', 'region': 'us-west-2'}},
 'status': {'ready': False, 'state': 'Initializing'}}]
>>> results
{'indexes': [{'dimension': 3,
              'host': 'jen3-sl-126ib2q.svc.dev-aws-ram-paid.pinecone.io',
              'metric': 'cosine',
              'name': 'jen3-sl',
              'spec': {'serverless': {'cloud': 'aws', 'region': 'us-west-2'}},
              'status': {'ready': True, 'state': 'Ready'}},
             {'dimension': 2,
              'host': 'jen2-sl-126ib2q.svc.apw5-4e34-81fa.pinecone.io',
              'metric': 'cosine',
              'name': 'jen2-sl',
              'spec': {'serverless': {'cloud': 'aws', 'region': 'us-west-2'}},
              'status': {'ready': False, 'state': 'Initializing'}}]}
```

## Type of Change

## Test Plan

Added unit tests


gdj0nes:
Can you just update the existing classes with __iter__ ?

jhamon:
> Can you just update the existing classes with `__iter__`? 
> 
> ```
> def __iter__(self):
>     return self.indexes 
> ```
> 
>  

No, because they are outputs of the generator and all models share the same mustache template.

tests/unit/models/test_iterable_index_list.py
```diff
@@ -0,0 +1,51 @@
...
+
+@pytest.fixture
+def index_list_response():
+    return IndexList(
+            indexes=[
+                {
+                    "name": "test-index-1",
+                    "dimension": 2,
+                    "metric": "cosine",
+                    "capacity_mode": "serverless",
```
austin-denoble, line 13: Do these index models need to be updated to the new `spec` shape? Could we use `IndexModel` directly like below in `test_control.py`?

austin-denoble (approved):
Very nice, wrapping these with some further utilities feels like the right thing to do here. Great work.

tests/unit/models/test_iterable_index_list.py
jhamon: Copilot let me down here 😂 

I'll update them. It doesn't really affect the way the tests work, but it makes the test better documentation if we use correct data.

PR #249, opened 2023-11-27
Update client with inlined type definitions under IndexModel and CreateIndexRequest

jhamon:
## Problem

We want to keep the amount of generated cruft being pushed into peoples' faces to a minimum

## Solution

We made some changes to our private openapi spec, then regenerated the core. Now fields such as name, dimension, and metric should be native python types.

## Usage

Screenshot 2023-11-26 at 11 11 27 PM

## Type of Change


austin-denoble (approved):
🚢

PR #251, opened 2023-11-29
Cleanup exception hierarchy

jhamon:
## Problem

We want to create a base exception `PineconeException` that all other errors extend to simplify error handling. 

## Solution

These changes are mainly derived from updated openapi templates. New exception hierarchy is:

- PineconeException
    - PineconeConfigurationError
    - PineconeApiException
        - NotFoundException
        - ForbiddenException
        - ...etc
    - PineconeProtocolError (GRPC)

Almost all find-and-replace changes except some tiny modifications in `pinecone/exceptions.py` to extend a different type with the same name in order for all types to share a common base type coming from the generated code. Not much to see here.

## Type of Change

### DIFF
--- a/pinecone/config/config.py
+++ b/pinecone/config/config.py
@@ -2,7 +2,7 @@ from typing import NamedTuple, Optional, Dict
 import os
 
 from pinecone.exceptions import PineconeConfigurationError
-from pinecone.core.client.exceptions import ApiKeyError
+from pinecone.core.client.exceptions import PineconeApiKeyError
 from pinecone.config.openapi import OpenApiConfigFactory
 from pinecone.core.client.configuration import Configuration as OpenApiConfiguration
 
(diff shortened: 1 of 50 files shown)


pinecone/config/config.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 from pinecone.exceptions import PineconeConfigurationError
-from pinecone.core.client.exceptions import ApiKeyError
+from pinecone.core.client.exceptions import PineconeApiKeyError
```
austin-denoble, line 5: Doesn't seem like `ApiKeyError` was used in this file previously. Is this import needed?

austin-denoble (approved):
Looks great, one question about an import that seems unused but I may be wrong. 🚢

pinecone/config/config.py
jhamon: Iffy. I can add it to the exports list I suppose in case anyone wants to catch that without drilling down into subpackages.

PR #252, opened 2023-11-29
[spruce] Add names() helper method on list results

jhamon:
## Problem

I'd like to make it even simpler to iterate over index names.

## Solution

With this change, you can now

```python
from pinecone import Pinecone

pc = Pinecone(api_key='xxx')
for name in pc.list_indexes().names():
  # do something
```
## Type of Change

## Test Plan

Unit test added.


austin-denoble (approved):
Nice one. 👍

PR #253, opened 2023-12-01
[spruce] Fix issue with polling after delete index

jhamon:
## Problem

There was a small error in a recent change to add a `names()` utility method on the response from `Pinecone#list_indexes` to facilitate easy iteration over index names. This resulted in an error when deleting indexes:

```
    def get_remaining():
>       return name in api_instance.list_indexes().names()
E       TypeError: 'NoneType' object is not callable
```

This error got introduced because `api_instance.list_indexes()` and `self.list_indexes()` are both a thing in this context, but only `self.list_indexes()` returns a result with the `.names()` helper.

## Solution

Call names on the correct object to avoid the error.

## Type of Change

## Test Plan

I confirmed with manual testing that this resolves the issue.

### DIFF
--- a/pinecone/control/pinecone.py
+++ b/pinecone/control/pinecone.py
@@ -123,7 +123,7 @@ class Pinecone:
         self.index_host_store.delete_host(self.config, name)
 
         def get_remaining():
-            return name in api_instance.list_indexes().names()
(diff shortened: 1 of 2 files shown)


austin-denoble (approved):
Nice catch. 🚢

PR #255, opened 2023-12-20
[Bug] Stop coercion of strings into datetime objects

jhamon:
## Problem

Sometimes when strings are stored as metadata values, later fetching or querying causes those values to be coerced into datetime objects by the client before they are presented to the caller. This is a default behavior in the generated openapi client that is a result of the python `datetime` constructor working quite hard to turn things into dates. We've seen scenarios where this was not desired, so for now we will disable this conversion process to have consistent handling of all strings. People who desire date objects can easily implement the conversion themselves.

## Solution

- Update openapi templates to comment out and remove logic related to converting strings into date and datetime objects.
- Regenerate client with new templates
- Add unit test verifying the type coercion does not take place. 

## Type of Change

## Test Plan

`poetry run pytest tests/unit/data`

### DIFF
--- a/pinecone/core/client/model_utils.py
+++ b/pinecone/core/client/model_utils.py
@@ -16,8 +16,6 @@ import pprint
 import re
 import tempfile
 
-from dateutil.parser import parse
-
 from pinecone.core.client.exceptions import (
     PineconeApiKeyError,
     PineconeApiAttributeError,
@@ -63,7 +61,7 @@ class cached_property(object):
             return result
 
 
-PRIMITIVE_TYPES = (list, float, int, bool, datetime, date, str, file_type)
+PRIMITIVE_TYPES = (list, float, int, bool, str, file_type)
 
 def allows_single_value_input(cls):
     """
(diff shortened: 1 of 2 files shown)


pinecone/core/client/model_utils.py
```diff
@@ -63,7 +61,7 @@ def __get__(self, instance, cls=None):
             return result
 
 
-PRIMITIVE_TYPES = (list, float, int, bool, datetime, date, str, file_type)
+PRIMITIVE_TYPES = (list, float, int, bool, str, file_type)
```
gdj0nes, line 64: Should we disable file type? Seems like a potential unnecessary conversion.

gdj0nes (approved):

austin-denoble (approved):
Looks good, thanks for fixing this one up. 👍 🚢

pinecone/core/client/model_utils.py
jhamon: Yeah, I can yank that out too easily enough.

jhamon: Actually, even though I think this should be safe to do, references to `file_type` are spread across a couple different files. So I will make a different ticket to handle that a bit later.

PR #258, opened 2023-12-26
Wire up pool_threads config

jhamon:
## Problem

We need to expose the `pool_threads` config in the new `Pinecone` constructor. This config property is ultimately used by the `ApiClient` class ([see here](https://github.com/pinecone-io/pinecone-python-client/blob/e81807de685d08297ea4ad642639fbc0d0ab24fd/pinecone/core/client/api_client.py#L61)), which is generated by the OpenAPI python generator. It is only used when performing operations over REST using the `async_req=True` param.

## Solution

Add a property and wire it up with the underlying `ApiClient` object and `Index` constructor.

## Type of Change

### DIFF
--- a/pinecone/control/pinecone.py
+++ b/pinecone/control/pinecone.py
@@ -24,6 +24,7 @@ class Pinecone:
         host: Optional[str] = None,
         config: Optional[Config] = None,
         additional_headers: Optional[Dict[str, str]] = {},
+        pool_threads: Optional[int] = 1,
         index_api: Optional[IndexOperationsApi] = None,
         **kwargs,
     ):
@@ -39,7 +40,8 @@ class Pinecone:
         if index_api:
             self.index_api = index_api
         else:
-            api_client = ApiClient(configuration=self.config.openapi_config)
+            self.pool_threads = pool_threads
+            api_client = ApiClient(configuration=self.config.openapi_config, pool_threads=pool_threads)
             api_client.user_agent = get_user_agent()
             extra_headers = self.config.additional_headers or {}
             for key, value in extra_headers.items():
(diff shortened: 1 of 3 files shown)


pinecone/control/pinecone.py
```diff
@@ -39,7 +40,8 @@ def __init__(
         if index_api:
             self.index_api = index_api
         else:
-            api_client = ApiClient(configuration=self.config.openapi_config)
+            self.pool_threads = pool_threads
+            api_client = ApiClient(configuration=self.config.openapi_config, pool_threads=pool_threads)
```
austin-denoble, line 44: tiny nit, probably irrelevant but should we pass in the class instance here similar to `configuration`?

```suggestion
            api_client = ApiClient(configuration=self.config.openapi_config, pool_threads=self.pool_threads)
```

pinecone/control/pinecone.py
```diff
@@ -222,6 +224,7 @@ def _get_status(self, name: str):
         response = api_instance.describe_index(name)
         return response["status"]
 
-    def Index(self, name: str):
-        index_host = self.index_host_store.get_host(self.index_api, self.config, name)
-        return Index(api_key=self.config.api_key, host=index_host)
+    def Index(self, name: str, host: Optional[str] = None):
```
austin-denoble, line 227: Tiny tweak for supporting custom `host` in grpc, right?

austin-denoble (approved):
Looks good. Just a couple questions / nit. 🚢

pinecone/control/pinecone.py
jhamon: Yes, this change is unrelated to `pool_threads` since that only applies to async REST calls, but I wanted to keep these method signatures aligned across both GRPC and non-GRPC usage. I made this change for the REST case so I wouldn't have to figure out how to stub out the call to fetch the host in the test that I added, and in general having more ways to avoid making that network call is a good thing.

PR #262, opened 2023-12-27
Update publish job with pypi api tokens

jhamon:
## Problem

Now that PyPI has required 2FA for all accounts, we have to publish using API Tokens instead of username and password.

## Solution

- Update workflows with pypi api token.
- Delete unused jobs that publish to testpypi

## Type of Change

## Test Plan


austin-denoble (approved):
👍 Thanks for fixing this up, Jen. The nightly spruce job was publishing, which is how I was alerted to things being busted, but I think you've been publishing dev builds directly rather than using that so it makes sense to clean those up.

PR #263, opened 2024-01-04
Dependency testing, support python3.8, downgrade urllib3

jhamon:
## Problem

Some goals:
- Verify our code works with the claimed dependency ranges
- Bring back python 3.8 support for LlamaIndex compatibility
- Add flexibility to urllib3 dependency version by extending the version range back to 1.26.x.

## Solution

- Remove more unused dependencies
    - `requests` (not sure when this was added; doesn't seem like it's been used for a while)
    - `python-dateutil` (no longer needed now that we've removed some date parsing default behavior in the generated client code) 
- Specify python-version specific dependency configuration for some dev dependencies (numpy, pandas-stubs) that were pushing us toward dropping py3.8 support
- Added new reusable workflow action, `.github/actions/create-index/`, to create an index that will be used for subsequent testing.
- Add new test workflow jobs, `dependency-matrix-rest` and `dependency-matrix-grpc`, with minimal integration test to verify the package works with different combinations of installed python dependencies.
  - New tests stored under `tests/dependency`  
  - Integration test exercises control plane (list indexes, describe an index ) and data plane (upsert and query). It does not create a new index for each test as this is time-consuming, the test matrix is large, and index creation is extensively tested elsewhere. 
  - Test configuration was getting unwieldy, so I split `testing.yaml` into `testing-rest.yaml` and `testing-grpc.yaml`
  - Data freshness isn't a huge issue for these tests, since the index is reused for upsert/query across different configurations. It's sufficient for these minimal tests to issue network calls and get responses without throwing exceptions to verify the dependencies are working.

## Type of Change

Is "being able to install" a feature?

## Test Plan

Add new test matrix to verify some dependencies work across the claimed range.


pyproject.toml
```diff
@@ -54,20 +54,36 @@ classifiers=[
...
+[tool.poetry.group.types.dependencies]
+mypy = "^1.6.1"
+types-urllib3 = "^1.26.25.14"
+grpc-stubs = "^1.53.0.3"
+pandas-stubs = [
+    {version = "^2.1.1.230928", python = ">=3.9"},
+    {version = "^1.5.3.230321", python = ">=3.8,<3.9"}
+]
+types-tqdm = "^4.66.0.3"
+types-protobuf = "^4.24.0.4"
```
jhamon, line 79: These are only used for running `mypy` type checks. I moved them into a different group hoping that would isolate them from the "prod" dependencies, but the poetry resolver requires all combinations of groups and extras to be compatible so it actually didn't meaningfully help solve my issue. In the end I had to add this additional version config for pandas-stubs in order to bring back python 3.8 support since the 2.1.1.x stubs were python >=3.9 only.

austin-denoble: Thanks for digging into all this, does not sound fun to have tracked down.

pyproject.toml
```diff
@@ -54,20 +54,36 @@ classifiers=[
...
+urllib3 = ">=1.26.16"
 tqdm = ">=4.64.1"
 grpcio = { version = ">=1.44.0", optional = true }
 grpc-gateway-protoc-gen-openapiv2 = { version = "0.1.0", optional = true }
 googleapis-common-protos = { version = ">=1.53.0", optional = true }
 lz4 = { version = ">=3.1.3", optional = true }
 protobuf = { version = "~=3.20.0", optional = true }
 
+[tool.poetry.group.types]
+optional = true
```
austin-denoble, line 68: Does this just flag everything in that group as optional rather than marking them explicitly?

.github/workflows/testing.yaml
```diff
@@ -133,3 +133,52 @@ jobs:
...
+          - 1.26.16
+          - 1.26.18
+          - 2.0.2
+          - 2.0.5
+          - 2.1.0
+          # - 1.26.17
+          # - 2.0.3
+          # - 2.0.4
+          # - 2.0.6
+          # - 2.0.7
```
austin-denoble, line 165: Can these comments be removed or is the intention to add them back in later?

.github/workflows/testing.yaml
```diff
@@ -87,7 +88,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9, '3.10', 3.11]
+        python-version: [3.9, '3.10', 3.11]
```
austin-denoble, line 91: I think I'm getting a bit mixed up on which python versions apply to where. Why do we remove 3.8 from here and not `units-no-grpc` above?

tests/dependency/grpc/test_sanity.py
```diff
@@ -0,0 +1,50 @@
...
+        
+        idx = client.Index(index_name)
+        idx.upsert(vectors=[
+            ('1', [1.0, 2.0]), 
+            ('2', [3.0, 4.0]),
+            ('3', [5.0, 6.0])
+        ])
+
+        # Wait for index freshness
+        time.sleep(30)
```
austin-denoble, line 40: Is grpc freshness a bit quicker than rest? We're using 30 seconds here and 60 for rest.

austin-denoble (approved):
Huge thank you for digging through dependency hell on this one. I have a few questions around python version in some of our CI testing, and some of the poetry stuff, but overall I think this makes sense.

pyproject.toml
jhamon: Poetry has "groups" for dev dependencies and "extras" for installable add-ons like "grpc". Similar concepts, but one is for us and one is for the end-users.

Groups are all included by default when you run `poetry install` but you can invert that behavior for specific groups by marking them optional and then explicitly controlling the groups you would like to install using `--with` and `--without` options. E.g. `poetry install --without dev` would install with nothing from the dev group (otherwise included by default), and `poetry install --with types` would install with the `dev` group (included by default) plus the `types` group (optional, so only included when specifically asked for).

.github/workflows/testing.yaml
jhamon: I haven't decided yet. Testing every single release kinda seems like overkill. The max jobs per workflow is 256 so I think I should keep it to this abbreviated list for now and save space for matrix runs on other jobs in the workflow.

.github/workflows/testing.yaml
jhamon: This was a mistake. At one point I was cleaning up 3.8 references in this workflow config because we'd partially removed it, but then later I figured out how to add it back and just overlooked this spot.

tests/dependency/grpc/test_sanity.py
jhamon: I don't think it's quicker. Just these aren't enabled in CI yet.

PR #264, opened 2024-01-05
Restore __version__ attribute to package

jhamon:
## Problem

`__version__` attribute accidentally got removed in an earlier refactor of the package, but it is still useful when code needs to check what client version they are dealing with.

## Solution

- Utils refactoring
  - Removed `__environment__` and associated code in `pinecone/utils/environment.py` utils, which actually do seem unnecessary. These are only referenced from another util file to define a constant that is also unused.
  - Migrated remaining code for reading version out of environment.py and into `pinecone/utils/version.py`. This reads a value from `pinecone/__version__` file. The value in this file is automatically updated during the publish workflow.
  - Update version-adjacent code involved in building the useragent string. Added some simple unit tests to verify these work as expected.
  - In general, move away from multiple usages of the `get_version()` function in favor of a cached value to avoid unnecessary disk reads.
- Exposed `__version__` attribute in top level `__init__.py`. This is what actually makes `pinecone.__version__` work.

### Future todo:
- The version number does get aliased to a constant called `CLIENT_VERSION` which seems silly, but I'll leave for now to avoid broadening the scope of this change even further to fix the usages of that constant. 

## Type of Change

## Test Plan


pinecone/utils/version.py
```diff
@@ -3,6 +3,4 @@
 def get_version():
     return Path(__file__).parent.parent.joinpath("__version__").read_text().strip()
 
-
-def get_environment():
-    return Path(__file__).parent.parent.joinpath("__environment__").read_text().strip()
\ No newline at end of file
+__version__ = get_version()
```
jhamon, line 6: There's really no reason for anyone outside this file to use `get_version()` since it just leads to reading from disk multiple times unnecessarily.

zackproser:
Thank you for tackling this so quickly!

austin-denoble (approved):
LGTM, nice cleanup. 🚢

PR #265, opened 2024-01-08
Accept host urls both with and without https prefix

jhamon:
## Problem

We want to make the client flexible enough to accept data plane host values both with and without the `https://` prefix. This is needed because the value is not prefixed with `https` in the describe_index response, but parts of the UI show the complete url with protocol.

## Solution

Add a new util method to normalize the urls. Incorporate the util into the IndexHostStore and ConfigBuilder. The ConfigBuilder usage should cover both the `Index` and `IndexGRPC` milestones.

## Type of Change

## Test Plan

Added new unit tests of util.


austin-denoble (approved):
LGTM, will get this added on the TS side today as well. 🚢

PR #268, opened 2024-01-10
Data plane integration testing suite

jhamon:
## Problem

Want to add a much more comprehensive integration testing suite against the data plane. This is a large undertaking because of the amount of variables involved, but this is a start.

Some variables to consider:
- REST vs GRPC
- pod vs serverless
- environment/cloud/region
- freshness timeout
- index metric, since hybrid search only makes sense for dotproduct

## Solution

Conceptually, I want every test run on github actions to target one combination of the factors above. This means some things which might otherwise be handled through test parameterization in a big monolithic test suite that is run with a single command are now toggled through environment variables. This makes it very easy to setup parallel runs with different configurations, but at the cost of some additional complexity. There's currently no single command to run every combo from my local machine in one go.

To run these locally right now, I might have an incantation like this:

```shell
PINECONE_API_KEY='nana-nana-nana-nana-batman' \
  METRIC='cosine' \
  USE_GRPC='false' \
  SPEC='{"serverless": {"region": "us-west-2", "cloud": "aws"}}' \
  poetry run pytest tests/integration/data/ -s -vv
``` 

- In CI, I shove this ugliness into a reusable github action `test-data-plane` that allow tests to easily be run using different configurations for target environment/cloud/region, REST vs GRPC, and metric type (which is relevant because hybrid search / sparse vector stuff only makes sense in the context of indexes with metric=dotproduct).
- Since waiting on data freshness is another bottleneck in the execution, I try to seed some data upfront and use it for all my query/fetch testing.
- By using the exact same tests via GRPC and REST for the first time, I'm discovering a lot of small details that don't match. So I'm making adjustments as I go to bring them into alignment, at least on the happy path. Error handling is a complex beast and will need more iteration in the future.
  - Migrated some logic from IndexGRPC in to another vector factory class to allow `IndexGRPC#upsert` to accept `Vector` or `VectorGRPC` args. This greatly simplifies test reuse and will be a nice UX improvement for users also because it simplifies docs.
  - Refactored a bit in the GRPC "unit" tests which were and still are a disaster. I found myself crawling through those after refactoring the `IndexGRPC` class to use the new `VectoryFactoryGRPC` class and triggering a few failures due to missed a few edge cases expressed through the unit tests. I split things up into smaller files and transitioned from using `self.<field>` everywhere to using pytest fixtures instead. It still sucks, but readability maybe slightly improved.
  - An example of something that changed (for alignment reasons) is that in the past an empty fetch result would be `None` via GRPC or `FetchResult(vectors=[])` via REST. The simplicity of `None` was nice, but now that there is `read_unit` information to display even when there are no results, the `FetchResponse` return value makes more sense.
- Implemented a vector factory for GRPC that is very similar to the one I created a while ago for the REST code path. It's much much easier to test edge cases in that class alone than in a more integrationy unit test that requires me to mock/patch a bunch of stuff that would otherwise fire off network calls.

To save resources/time while initially building out these tests, for now I'm only running with serverless indexes. But it should be easy to add other configurations soon.

## Some possible bugs/questions to follow up:
- Seems like `$ne` and `$nin` aren't working [as documented](https://docs.pinecone.io/docs/metadata-filtering)
- Freshness seems much different on dotproduct indexes. Not sure if this is expected.
- Need to rename `readUnits` to `read_units` before shipping 3.0

## High-level coverage so far

- [x] Upsert
- [x] Upsert (sparse vec)
- [x] Fetch
- [x] Query
- [x] Query (metadata filters)
- [ ] Query (hybrid / sparse vec)
- [ ] Update
- [ ] DescribeIndexStats
- [ ] Delete

## Type of Change


.github/actions/test-data-plane/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+    required: false
+    default: '60'
+  PINECONE_API_KEY:
+    description: 'The Pinecone API key'
+    required: true
+
+outputs:
+  index_name:
+    description: 'The name of the index, including randomized suffix'
+    value: ${{ steps.create-index.outputs.index_name }}
```
austin-denoble, line 32: Is this referring to a different `create-index` step that's present in this action, or is it assuming it's been run prior? It's working properly so I'm just going to assume the scope for `steps.<whatever>` is global to the whole workflow. 😅

.github/workflows/testing-integration.yaml
```diff
@@ -0,0 +1,91 @@
...
+    name: Data plane serverless integration tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        use_grpc: [true, false]
+        metric: 
+          - cosine
+          # - euclidean
+          # - dotproduct
```
austin-denoble, line 16: Assuming you want to keep these commented out for now, but if not.

.github/workflows/testing-dependency.yaml
```diff
@@ -124,5 +94,49 @@ jobs:
...
+    needs: dependency-matrix-setup
+    strategy:
+      max-parallel: 5
+      fail-fast: false
+      matrix:
+        python-version:
+          - 3.8
+          - 3.9
+          - '3.10'
+          - 3.11
```
austin-denoble, line 114: Everything but `3.8` is commented out in `dependency-matrix-grpc`, is that intentional?

pinecone/grpc/vector_factory_grpc.py
```diff
@@ -0,0 +1,122 @@
...
+    Vector as GRPCVector,
+    QueryVector as GRPCQueryVector,
+    SparseValues as GRPCSparseValues,
+)
+from pinecone import (
+    Vector as NonGRPCVector, 
+    SparseValues as NonGRPCSparseValues
+)
+
+class VectorFactoryGRPC:
```
austin-denoble, line 31: Thanks for all the work getting the grpc side of things a bit more aligned with the non-grpc paths. 🙏

pinecone/grpc/vector_factory_grpc.py
```diff
@@ -0,0 +1,122 @@
...
+    @staticmethod
+    def build(item: Union[GRPCVector, NonGRPCVector, Tuple, Dict]) -> GRPCVector:
+        if isinstance(item, GRPCVector):
+            return item
+        elif isinstance(item, NonGRPCVector):
```
austin-denoble, line 36: I don't think it would, but would the `VectorFactory` benefit from having the flexibility of `GRPCVector` being added to the union in `build`?

tests/integration/data/conftest.py
```diff
@@ -0,0 +1,83 @@
...
+    return json.loads(get_environment_var('SPEC'))
+
+@pytest.fixture(scope='session')
+def index_name():
+    # return 'dataplane-lol'
+    return 'dataplane-' + random_string(20)
+    
+@pytest.fixture(scope='session')
+def namespace():
+    # return 'banana'
```
austin-denoble, line 49: Assuming these return comments can come out.

tests/integration/data/conftest.py
```diff
@@ -0,0 +1,83 @@
...
+    return random_string(10)
+
+@pytest.fixture(scope='session')
+def idx(client, index_name, index_host):
```
austin-denoble, line 53: **nitpick:** Any specific reason for using `idx`?

tests/integration/helpers/helpers.py
```diff
@@ -44,3 +45,41 @@ def get_environment_var(name: str, defaultVal: Any = None) -> str:
...
+        stats = idx.describe_index_stats()
+        if namespace in stats.namespaces and stats.namespaces[namespace].vector_count >= expected_count:
+            done = True
+        elif total_time > max_sleep:
+            raise TimeoutError(f'Timed out waiting for namespace {namespace} to have vectors')
+        else:
+            total_time += delta_t
+            time.sleep(delta_t)
+
+def poll_fetch_for_ids_in_namespace(idx, ids, namespace):
```
austin-denoble, line 64: 👍 Lots of goodies for improving thinking around how to streamline the integration flows on typscript side, thank you again.

tests/unit/utils/test_convert_to_list.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 from pinecone.utils import convert_to_list
+from pinecone import SparseValues
```
austin-denoble, line 3: **nitpick:** Might not be needed.

tests/unit_grpc/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    return [0.7, 0.3, 0.31415]
+
+
+@pytest.fixture
+def md1():
+    return {"genre": "action", "year": 2021}
+
+
+@pytest.fixture
+def md2():
```
austin-denoble, line 44: **nitpick:** More descriptive `metadata1()` and `metadata2()`? Very nitpicky, but since these fixtures are used across tests in other files might help readability.

tests/unit_grpc/test_grpc_index_describe_index_stats.py
```diff
@@ -0,0 +1,37 @@
...
+)
+from pinecone.grpc.utils import dict_to_proto_struct
+
+
+class TestGrpcIndexDescribeIndexStats:
+    def setup_method(self):
+        self.config = Config(api_key="test-api-key", host="foo")
+        self.index = GRPCIndex(config=self.config, index_name="example-name", _endpoint_override="test-endpoint")
+
+    def test_describeIndexStats_callWithoutFilter_CalledWithoutFilter(self, mocker):
```
austin-denoble, line 21: **nitpick:** I was going to ask about the duplicative seeming naming for tests in the grpc suite, but it seems to be across the board in some of these files.

austin-denoble (approved):
Sorry for all the nitpicks and questions, you can obviously ignore and land as-is. This is fantastic, thanks for expanding our coverage, streamlining the integration flows, and packing in additional GRPC quality of life improvements. Also hot on the heels of all the dependency coverage work, wow! 🚢

.github/actions/test-data-plane/action.yaml
jhamon: I think this `outputs` declaration is just accidental copy/pasta from the `create-index` action I made earlier this week for dependency testing purposes so it can be removed. If I wanted to use it I could by having my test suite write the name of the index it creates to the `GITHUB_OUTPUT` location (see [here](https://github.com/pinecone-io/pinecone-python-client/blob/spruce/scripts/create.py#L15) in script used by create-index action) but in this case it's not needed since the test creates and deletes the index.

.github/workflows/testing-dependency.yaml
jhamon: Eh, not exactly. Mainly just trying to save some time and resources while iterating. I should bring them back on the grpc job.

.github/workflows/testing-integration.yaml
jhamon: Yeah, leaving them out while still fleshing out the tests and iterating heavily. Dotproduct is having some kind of performance issue at the moment the db team are looking into.

pinecone/grpc/vector_factory_grpc.py
jhamon: Perhaps in some circumstances, but it's probably a lot less likely to happen than people trying to pass the `NonGRPCVector` object into into the GRPC-flavored client. The GRPC-flavored version of this object is annoying because you have to convert metadata dicts yourself into `Struct` before passing the kwarg param. So the ergonomics leave a lot to be desired.

tests/unit_grpc/conftest.py
jhamon: I find everything about these tests loathsome even after some refactoring. 💀 

They were previously using attributes on a test class called `self.md1`, etc, so I kept the names the same for now just to simplify the refactor.

PR #270, opened 2024-01-12
Add `usage.read_units` in fetch/query responses

jhamon:
## Problem

Need to update the generated core and surface usage data properly.

## Solution

- Make a few adjustments related to tag names in updated OpenAPISpec
- Update data integration tests to assert `read_units` are present
- Make adjustments to GRPC response parsing to account for the new data attributes

## Type of Change


pinecone/grpc/utils.py
```diff
@@ -52,13 +54,19 @@ def parse_fetch_response(response: dict):
...
+    usage = parse_usage(response)
     
     return FetchResponse(
         vectors=vd, 
         namespace=namespace,
+        usage=usage,
         _check_type=False
     )
 
+def parse_usage(response):
```
austin-denoble, line 67: Is the parsing still needed if it's an int, or is this grpc-specific?

austin-denoble (approved):
LGTM 🚢

PR #272, opened 2024-01-13
Remove deprecated `queries` option in `query()`

jhamon:
## Problem

There was an argument on the `query()` method for passing multiple queries at once. This has long been deprecated in the API and SDK, so it's time to remove it.
 
## Solution

- Clean up the `queries` option to both `Index#query` and `IndexGRPC#query`.
- Remove code no longer needed to interpret the many forms these query options could take
- Cleanup and simplify response parsing logic
- Remove related tests passing `queries` kwarg

## Type of Change

## Test Plan

Recently added integration test suite gives high confidence queries still work after this change.

### DIFF
--- a/pinecone/data/index.py
+++ b/pinecone/data/index.py
@@ -11,7 +11,6 @@ from pinecone.core.client.models import (
     FetchResponse,
     QueryRequest,
     QueryResponse,
-    QueryVector,
     RpcStatus,
     ScoredVector,
     SingleQueryResults,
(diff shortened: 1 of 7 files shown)


pinecone/data/index.py
```diff
@@ -371,35 +361,19 @@ def query(
...
                              {'indices': List[int], 'values': List[float]}, where the lists each have the same length.
-        Keyword Args:
-            Supports OpenAPI client keyword arguments. See pinecone.core.client.models.QueryRequest for more details.
-
+        
         Returns: QueryResponse object which contains the list of the closest vectors as ScoredVector objects,
                  and namespace name.
         """
 
-        def _query_transform(item):
```
austin-denoble, line 381: 🗑️

austin-denoble (approved):
🚢

PR #278, opened 2024-01-14
Add missing .names() for list_collections()

jhamon:
## Problem

The `list_indexes()` response has a convenience method that lets you easily get index names by chaining `.names()`. We want the same ergonomics on `list_collections()`.

## Solution

Add test and adjust model for implementation. Now you should be able to do:

```python
if index_name not in pc.list_indexes.names():
  # this already was working

if collection_name not in pc.list_collections.names():
  # this is what got added in this diff
```

## Type of Change


austin-denoble (approved):
🚢

PR #280, opened 2024-01-15
Simplify passing host configuration direct to Index client

jhamon:
## Problem

In production situations, it's important people be able to configure the client in a way that does not hit the control plane to find out the host url. This was possible, but now that I'm writing docs I realized this is cumbersome because the positional index name arg was still being required even though it wasn't being used in that scenario. 

We want it to be easy.

### Before

```python
from pinecone.grpc import PineconeGRPC

pc = PineconeGRPC(api_key="key")

# Targeting by name is easy
index = pc.Index('blah')

# Targeting by host url is awkward
index = pc.Index('required-but-unused-index-name', host='blah-24vnhz6.svc.apw5-4e34-81fa.pinecone.io')
```

### After

```python
from pinecone.grpc import PineconeGRPC

pc = PineconeGRPC(api_key="key")

# This still works
index = pc.Index('blah')

# This now works
index = pc.Index(host='blah-24vnhz6.svc.apw5-4e34-81fa.pinecone.io')

# Or pass both if you really want, but not needed.
index = pc.Index(name='blah', host='blah-24vnhz6.svc.apw5-4e34-81fa.pinecone.io')
```

## Type of Change

## Test Plan

Update integration tests to check it can still issue data calls without errors no matter how it is configured.

### DIFF
--- a/pinecone/control/pinecone.py
+++ b/pinecone/control/pinecone.py
@@ -7,7 +7,7 @@ from pinecone.config import PineconeConfig, Config
 
 from pinecone.core.client.api.manage_indexes_api import ManageIndexesApi as IndexOperationsApi
 from pinecone.core.client.api_client import ApiClient
-from pinecone.utils import get_user_agent
+from pinecone.utils import get_user_agent, normalize_host
 from pinecone.core.client.models import (
     CreateCollectionRequest,
     CreateIndexRequest,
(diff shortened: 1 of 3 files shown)


tdonia (approved):
💜

austin-denoble (approved):
Nice! 🚢

PR #285, opened 2024-01-17
Add `query()` validation to prevent sending both `id` and `vector`

jhamon:
## Problem

Currently the python SDK allows users to send a request that contains values for both the `id` and `vector` fields. This request doesn't make sense to the API since the `id` field, when provided, is used to look up a vector in the index whose values should be used as the query.

We'd like to give clear feedback to the user that this is not correct usage rather than showing empty results.

## Solution

- Implement the validation as a simple if-statement. 
- Update docstrings.

## Type of Change

- [x] None of the above: UX improvement

## Test Plan

Added unit tests for this error handling case

### DIFF
--- a/pinecone/data/index.py
+++ b/pinecone/data/index.py
@@ -368,6 +368,9 @@ class Index():
 
         _check_type = kwargs.pop("_check_type", False)
 
+        if vector is not None and id is not None:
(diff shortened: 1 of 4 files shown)


tdonia (approved):
:shipit:

PR #291, opened 2024-01-19
Improve `AttributeError` warnings when using deprecated methods

jhamon:
## Problem

In v3.0, many methods have moved from the top-level namespace onto a class instance. We can throw more informative errors to help people migrate.

## Solution

Add back stub implementations of methods that were removed. When called, they throw errors with useful documentation.

I simplified the expected args for all methods to `*args, **kwargs` which should match no matter what combination of arguments the user has provided.

## Type of Change

## Test Plan

`poetry run python3` and then try:

```
import pinecone

pinecone.list_indexes() # etc
```

### DIFF
--- a/pinecone/deprecation_warnings.py
+++ b/pinecone/deprecation_warnings.py
@@ -1,0 +1,123 @@
+def _build_class_migration_message(method_name: str, example: str):
+    return f"""
+The {method_name} method is no longer top-level attribute of the pinecone package.
+
+To use this method, please create a client instance and call the method.
+
+Example:
+{example}
+"""
+
+def list_indexes(*args, **kwargs):
+    example = """
+    from pinecone import Pinecone
+    
+    pc = Pinecone(api_key='YOUR_API_KEY')
+
+    index_name = "quickstart" # or your index name
+
+    if index_name not in pc.list_indexes().names():
+        # do something
+"""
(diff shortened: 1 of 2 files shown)


pinecone/deprecation_warnings.py
```diff
@@ -0,0 +1,123 @@
+def _build_class_migration_message(method_name: str, example: str):
+    return f"""
+The {method_name} method is no longer top-level attribute of the pinecone package.
```
austin-denoble, line 3: **nit:** ...is no longer **a** top-level attribute...

pinecone/deprecation_warnings.py
```diff
@@ -0,0 +1,123 @@
...
+def scale_index(*args, **kwargs):
+    example = """
+    from pinecone import Pinecone
+
+    pc = Pinecone(api_key='YOUR_API_KEY')
+    pc.configure_index('my_index', replicas=2)
+"""
+
+    msg = f"""
+The scale_index method is no longer top-level attribute of the pinecone package.
```
austin-denoble, line 69: **nit:** ...is no longer **a** top-level attribute...

austin-denoble (approved):
Nice one, looks good to me. 🚢

jseldess (approved):
LGTM

PR #296, opened 2024-01-22
Expose missing configurations for `grpc_config` and `pool_threads`

jhamon:
## Problem

This PR addresses a few configuration-related issues:
- We moved `pool_threads` configuration up to the `Pinecone` client instance, but for backwards compatibility it would be nice to also allow it to be passed in the call to `pc.Index(host='your-host', pool_threads=5)`
- We need to expose `grpc_config` as a keyword param when targeting indexes via the grpc-flavored client.

## Solution

- Pass through kwargs from `.Index()` to underlying Index class implementation
- Add unit tests making sure `pool_threads` and `grpc_config` handled correctly.

### Before (pool_threads)

```python
from pinecone import Pinecone

pc = Pinecone(api_key='key', pool_threads=5)
pc.Index(host='host')
pc.upsert(...)
```

### After (pool_threads, add backwards compatibility with old way)

You can use syntax as before, or alternatively can do this:

```python
from pinecone import Pinecone

pc = Pinecone(api_key='key')
index = pc.Index(host='host', pool_threads=5)
index.upsert(...)
```

### After (grpc, allow grpc_config to be passed)

```python
from pinecone.grpc import PineconeGRPC, GRPCClientConfig

pc = PineconeGRPC(api_key='YOUR_API_KEY')
grpc_config = GRPCClientConfig(
    timeout=10, 
    secure=True,
    reuse_channel=True
)
index = pc.Index(name='my-index', host='host', grpc_config=grpc_config)
```

## Type of Change

- [x] None of the above: Add backwards compatibility on pool_threads config

## Test Plan

New tests should be green in CI.

### DIFF
--- a/pinecone/control/pinecone.py
+++ b/pinecone/control/pinecone.py
@@ -94,10 +94,10 @@ class Pinecone:
         else:
             self.config = PineconeConfig.build(api_key=api_key, host=host, additional_headers=additional_headers, **kwargs)
 
+        self.pool_threads = pool_threads
         if index_api:
             self.index_api = index_api
         else:
-            self.pool_threads = pool_threads
             api_client = ApiClient(configuration=self.config.openapi_config, pool_threads=self.pool_threads)
             api_client.user_agent = get_user_agent()
             extra_headers = self.config.additional_headers or {}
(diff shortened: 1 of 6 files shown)


austin-denoble (approved):
LGTM 🚢

PR #297, opened 2024-01-23
Optional configs to pass `additional_headers`/`additional_metadata` to indexes

jhamon:
## Problem

There are some debugging scenarios where, in order to debug or track a specific request, we would like the ability to pass additional headers with each request.

## Solution

- Add `additional_headers` kwarg to `Index()` for REST calls
- Add `additional_metadata` field within existing `grpc_config` object for configuring equivalent for GRPC
- Add unit tests for both ways of doing it. Metadata is a [similar concept](https://grpc.io/docs/guides/metadata/) to http request headers in GRPC, not to be confused with vector metadata.

## Type of Change

- [x] None of the above: New feature, but really for Pinecone developer/support use only.

### Usage (REST)

```python
from pinecone import Pinecone

pc = Pinecone(api_key='xxx')
index = pc.Index(
    host='hosturl', 
    additional_headers={ 'header-1': 'header-1-value' }
)

# Now do things
index.upsert(...)
```

### Usage (GRPC)

```python
from pinecone.grpc import PineconeGRPC, GRPCClientConfig

pc = PineconeGRPC(api_key='YOUR_API_KEY')

grpc_config = GRPCClientConfig(additional_metadata={'extra-header': 'value123'})
index = pc.Index(
    name='my-index', 
    host='host', 
    grpc_config=grpc_config
)

# do stuff
index.upsert(...)
```

## Test Plan

Besides unit tests, I will try running some test commands with `PINECONE_DEBUG_CURL='true'` enabled to see what request is being sent.

### DIFF
--- a/pinecone/data/index.py
+++ b/pinecone/data/index.py
@@ -69,15 +69,25 @@ class Index():
     For improved performance, use the Pinecone GRPC index client.
     """
 
-    def __init__(self, api_key: str, host: str, pool_threads=1, **kwargs):
-        api_key = api_key or kwargs.get("api_key", None)
-        host = host or kwargs.get('host', None)
-        pool_threads = pool_threads or kwargs.get("pool_threads")
-
+    def __init__(
+            self,
+            api_key: str,
+            host: str,
+            pool_threads: Optional[int] = 1,
+            additional_headers: Optional[Dict[str, str]] = {},
+            **kwargs
+        ):
         self._config = ConfigBuilder.build(api_key=api_key, host=host, **kwargs)
         
-        api_client = ApiClient(configuration=self._config.openapi_config, pool_threads=pool_threads)
+        api_client = ApiClient(configuration=self._config.openapi_config, 
+                               pool_threads=pool_threads)
+
+        # Configure request headers
         api_client.user_agent = get_user_agent()
+        extra_headers = additional_headers or {}
+        for key, value in extra_headers.items():
+            api_client.set_default_header(key, value)
+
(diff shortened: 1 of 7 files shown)


pinecone/data/index.py
```diff
@@ -69,15 +69,25 @@ class Index():
     For improved performance, use the Pinecone GRPC index client.
     """
 
-    def __init__(self, api_key: str, host: str, pool_threads=1, **kwargs):
-        api_key = api_key or kwargs.get("api_key", None)
-        host = host or kwargs.get('host', None)
-        pool_threads = pool_threads or kwargs.get("pool_threads")
```
austin-denoble, line 75: Were these removed because we don't necessarily need to check for them in `kwargs` anymore?

austin-denoble (approved):
LGTM, tested locally and we've got headers properly applied to data plane operations:
Screenshot 2024-01-24 at 11 41 22 AM

pinecone/data/index.py
jhamon: I realized that the code wasn't doing anything. If you have a function that takes `**kwargs`, only unnamed keyword args that are not explicitly listed are present in the kwargs dictionary.

```python
>>> def see_kwargs(api_key, **kwargs):
...   print(kwargs)
...
>>> see_kwargs(api_key='xxx', extra_stuff=True)
{'extra_stuff': True}
>>> see_kwargs('xxx', more_extra=True)
{'more_extra': True}
```

PR #299, opened 2024-01-23
Integration tests for collections

jhamon:
## Problem

We received a bug report that creation of indexes using `PodSpec` fails if `source_collection` is specified.

## Solution

- The fix for the bug was a one-line change.
- Added several integration tests to exercise `index --> collection --> index` path and error cases.
- Restructured integration tests so that tests using pod-based indexes reside in `tests/integration/control/pod` and can be run separately from severless indexes tested in `tests/integration/control/serverless`. This allows for greater parallelism in CI.
- Adjusted CI configs to run these tests in parallel to integration tests using serverless indexes. The collections tests are quite slow due to the waiting required for pod indexes and collections to become ready for use.

## Type of Change


tests/integration/control/pod/test_collections_errors.py
```diff
@@ -0,0 +1,90 @@
...
+                dimension=dimension + 1, 
+                metric=metric, 
+                spec=PodSpec(
+                    environment=environment, 
+                    source_collection=reusable_collection
+                )
+            )
+        assert 'Index and collection must have the same dimension' in str(e.value)
+
+    # def test_create_index_from_notready_collection(self, client, ready_index, random_vector, dimension, metric, environment):
```
jhamon, line 56: Not sure about this scenario. Need to follow up later.

pinecone/models/pod_spec.py
```diff
@@ -60,6 +60,11 @@ class PodSpec(NamedTuple):
     {'indexed': ['field1', 'field2']}
     """
 
+    source_collection: Optional[str] = None
+    """
+    The name of the collection to use as the source for the pod index. This configuration is only used when creating a pod index from an existing collection.
+    """
+
```
jhamon, line 67: This missing property was the cause create index with `source_collection` was failing.

tests/integration/control/pod/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+            print(f"Attempting delete of index {index_name}")
+            client.delete_index(index_name, -1)
+            print(f"Deleted index {index_name}")
+            break
+        except Exception as e:
+            print(f"Unable to delete index {index_name}: {e}")
+            pass
+
+    if time_waited >= 120:
+        raise Exception(f"Index {index_name} is not ready to delete after 120 seconds")
```
austin-denoble, line 149: **nit:** Wording here could be: "Index {index_name} was not able to be deleted after 120 seconds"

tests/integration/control/pod/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+def environment():
+    return get_environment_var('PINECONE_ENVIRONMENT')
+
+@pytest.fixture()
+def dimension():
+    return int(get_environment_var('DIMENSION'))
+
+@pytest.fixture()
+def create_index_params(index_name, environment, dimension, metric):
+    spec = {
```
austin-denoble, line 26: Interesting, I thought you had to use the `PodSpec` class for `spec` rather than passing a plain object.

tests/integration/control/pod/test_configure_pod_index.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+import time
+
+class TestConfigurePodIndex():
+    def test_configure_pod_index(self, client, ready_index):
+        time.sleep(10) # Wait a little more, just in case.
+        client.configure_index(ready_index, replicas=1, pod_type='p1.x1')
```
austin-denoble, line 7: Do we need some asserts here?

austin-denoble (approved):
Nice work continuing to expand our testing setup. Some nits and questions, but this looks good to go. 🚢

tests/integration/control/pod/test_configure_pod_index.py
jhamon: Maybe someday. To start with, I just wanted to see if the request can be made without errors and this was enough to catch the bug we saw ahead of the release last week. Asserting what the server actually does as a result can come later, or possibly be covered in tests owned by platform.

PR #304, opened 2024-02-05
Support overriding `additional_headers` with `PINECONE_ADDITIONAL_HEADERS` environment variable

fsxfreak:
## Problem

We need a way to attach special request headers in integration testing.

## Solution

Add support for overriding additional_headers via the `PINECONE_ADDITIONAL_HEADERS` environment variable.

## Type of Change

## Test Plan

Unit tests

```
$ cd pinecone-python-client/
$ export PINECONE_ADDITIONAL_HEADERS='{"header": "value"}'
$ python3
>>> client = Pinecone(api_key='key', host='host')
>>> vars(client)
{'config': ... additional_headers={'header': 'value'}), ...}
```

### DIFF
--- a/pinecone/config/pinecone_config.py
+++ b/pinecone/config/pinecone_config.py
@@ -13,11 +13,12 @@ class PineconeConfig():
     @staticmethod
     def build(api_key: Optional[str] = None, host: Optional[str] = None, additional_headers: Optional[Dict[str, str]] = {},  **kwargs) -> Config:
         host = host or kwargs.get("host") or os.getenv("PINECONE_CONTROLLER_HOST") or DEFAULT_CONTROLLER_HOST
-        try:
-          headers_json = os.getenv("PINECONE_ADDITIONAL_HEADERS")
-          headers = json.loads(headers_json)
-          additional_headers = additional_headers or headers
-        except json.decoder.JSONDecodeError as e:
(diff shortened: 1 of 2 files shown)


pinecone/config/pinecone_config.py
```diff
@@ -1,12 +1,24 @@
...
 class PineconeConfig():
     @staticmethod
     def build(api_key: Optional[str] = None, host: Optional[str] = None, additional_headers: Optional[Dict[str, str]] = {},  **kwargs) -> Config:
         host = host or kwargs.get("host") or os.getenv("PINECONE_CONTROLLER_HOST") or DEFAULT_CONTROLLER_HOST
+        headers_json = os.getenv("PINECONE_ADDITIONAL_HEADERS")
+        if headers_json:
+          try:
+            headers = json.loads(headers_json)
+            additional_headers = additional_headers or headers
+          except json.decoder.JSONDecodeError as e:
```
jhamon, line 21: I would probably catch all exceptions here just to be robust against any changes that might occur in how the `json` package structures their errors. We have to work with multiple python versions and future versions that could make changes here.

```suggestion
          except Exception as e:
```

jhamon (approved):
Looks good to me.

pinecone/config/pinecone_config.py
fsxfreak: makes sense, will make the change

PR #307, opened 2024-02-06
List vector ids by prefix

jhamon:
## Problem

Need to implement the new data plane list endpoint.

## Solution

- Update generated code. Pull in list endpoint. All changes under `pinecone/core` are generated from spec files and can be ignored for this review.
- Implement changes for list endpoint in:
  - `pinecone/data/index.py` main implementation
  - `pinecone/grpc/index_grpc.py` main implementation
  - `tests/integration/data/conftest.py` adjustments to test setup, mainly to generate a new namespace to hold vectors for list testing data.
  - `tests/integration/data/seed.py` to upsert a larger number of vectors, so I would have enough data to page through
  - `tests/integration/data/test_list.py`
  -  `tests/integration/data/test_list_errors.py`

## Open questions

- Do we expect to ever return more data than just `{'id': '1'}` in the vectors array? For convenience the `list()` method is currently implemented as a generator function that abstracts the pagination steps and yields a flat list of id values. For a use case where you were going to immediately fetch those ids, this seems ideal. But would be limiting if we ever wanted to return more than just ids here.
 
## Usage

Install the dev client version `install "pinecone-client[grpc]"==3.1.0.dev1`

### REST

```python
from pinecone import Pinecone

pc = Pinecone(api_key='xxx')
index = pc.Index(host='hosturl')

# To iterate over all result pages using a generator function
for ids in index.list(prefix='pref', limit=3, namespace='foo'):
    print(ids) // ['pref1', 'pref2', 'pref3']

# For manual control over pagination
results = index.list_paginated(
    prefix='pref', 
    limit=3, 
    namespace='foo', 
    pagination_token='eyJza2lwX3Bhc3QiOiI5IiwicHJlZml4IjpudWxsfQ=='
)
print(results.namespace)
print([v.id for v in results.vectors])
print(results.pagination.next)
print(results.usage)
```

### GRPC

```python
from pinecone.grpc import PineconeGRPC

pc = PineconeGRPC(api_key='xxx')
index = pc.Index(host='hosturl')

# To iterate over all result pages using a generator function
for ids in index.list(prefix='pref', limit=3, namespace='foo'):
    print(ids) // ['pref1', 'pref2', 'pref3']

# For manual control over pagination
results = index.list_paginated(
    prefix='pref', 
    limit=3, 
    namespace='foo', 
    pagination_token='eyJza2lwX3Bhc3QiOiI5IiwicHJlZml4IjpudWxsfQ=='
)
print(results.namespace)
print([v.id for v in results.vectors])
print(results.pagination.next)
print(results.usage)
```

## Type of Change

## Testing

Try out the dev version.

```
pip install "pinecone-client[grpc]"==3.1.0.dev1
```

### DIFF
--- a/pinecone/data/index.py
+++ b/pinecone/data/index.py
@@ -512,6 +512,31 @@ class Index():
         namespace: Optional[str] = None,
         **kwargs
     ) ->  ListResponse:
+        """
+        The list_paginated operation finds vectors based on an id prefix within a single namespace.
+        It returns matching ids in a paginated form, with a pagination token to fetch the next page of results.
+        This id list can then be passed to fetch or delete operations, depending on your use case.
+        
+        Consider using the `list` method to avoid having to handle pagination tokens manually.
+
+        Examples:
+            >>> results = index.list_paginated(prefix='99', limit=5, namespace='my_namespace')
+            >>> [v.id for v in results.vectors]
+            ['99', '990', '991', '992', '993']
+            >>> results.pagination.next
+            eyJza2lwX3Bhc3QiOiI5OTMiLCJwcmVmaXgiOiI5OSJ9
+            >>> next_results = index.list_paginated(prefix='99', limit=5, namespace='my_namespace', pagination_token=results.pagination.next)
+
+        Args:
+            prefix (Optional[str]): The id prefix to match. If unspecified, an empty string prefix will 
+                                    be used with the effect of listing all ids in a namespace [optional]
+            limit (Optional[int]): The maximum number of ids to return. If unspecified, the server will use a default value. [optional]
+            pagination_token (Optional[str]): A token needed to fetch the next page of results. This token is returned 
+                in the response if additional results are available. [optional]
+            namespace (Optional[str]): The namespace to fetch vectors from. If not specified, the default namespace is used. [optional]
+        
+        Returns: ListResponse object which contains the list of ids, the namespace name, pagination information, and usage showing the number of read_units consumed.
+        """
         args_dict = self._parse_non_empty_args(
             [
                 ("prefix", prefix),
@@ -524,15 +549,33 @@ class Index():
 
     @validate_and_convert_errors
     def list(self, **kwargs):
-        limit = kwargs.get("limit", 100)
+        """
+        The list operation accepts all of the same arguments as list_paginated, and returns a generator that yields
+        a list of the matching vector ids in each page of results. It automatically handles pagination tokens on your
+        behalf.
+
+        Examples:
+            >>> for ids in index.list(prefix='99', limit=5, namespace='my_namespace'):
+            >>>     print(ids)
+            ['99', '990', '991', '992', '993']
+            ['994', '995', '996', '997', '998']
+            ['999']
+
+        Args:
+            prefix (Optional[str]): The id prefix to match. If unspecified, an empty string prefix will 
+                                    be used with the effect of listing all ids in a namespace [optional]
+            limit (Optional[int]): The maximum number of ids to return. If unspecified, the server will use a default value. [optional]
+            pagination_token (Optional[str]): A token needed to fetch the next page of results. This token is returned 
+                in the response if additional results are available. [optional]
+            namespace (Optional[str]): The namespace to fetch vectors from. If not specified, the default namespace is used. [optional]
+        """
         done = False
         while not done:
             results = self.list_paginated(**kwargs)
             if len(results.vectors) > 0:
                 yield [v.id for v in results.vectors]
             
-            full_page = len(results.vectors) == limit
-            if results.pagination and full_page:
+            if results.pagination:
                 kwargs.update({"pagination_token": results.pagination.next})
             else:
                 done = True
--- a/pinecone/data/index.py
+++ b/pinecone/data/index.py
@@ -1,6 +1,5 @@
 from tqdm.autonotebook import tqdm
 
-from collections.abc import Iterable
 from typing import Union, List, Tuple, Optional, Dict, Any
 
 from pinecone.config import ConfigBuilder
@@ -22,9 +21,10 @@ from pinecone.core.client.models import (
     DeleteRequest,
     UpdateRequest,
     DescribeIndexStatsRequest,
+    ListResponse
 )
 from pinecone.core.client.api.data_plane_api import DataPlaneApi
-from ..utils import get_user_agent, fix_tuple_length
+from ..utils import get_user_agent
 from .vector_factory import VectorFactory
 
 __all__ = [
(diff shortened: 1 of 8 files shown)


opbenesh:
@jhamon, awesome work. Excited to get that out in the open!

As for your open question: we do not plan to add anything else _in the near future_, but it theoretically might happen at some point. It's possible that if we add some _auxiliary_ data it'll be possible to modify the iterator function to deal with it/strip it away, though.

@gdj0nes WDYT?

pinecone/data/index.py
```diff
@@ -502,6 +502,40 @@ def describe_index_stats(
...
+    def list(self, **kwargs):
+        limit = kwargs.get("limit", 100)
+        done = False
+        while not done:
+            results = self.list_paginated(**kwargs)
+            if len(results.vectors) > 0:
+                yield [v.id for v in results.vectors]
+            
+            full_page = len(results.vectors) == limit
+            if results.pagination and full_page:
```
opbenesh, line 535: @jhamon it's valid for a page to contain less than `limit` results without being the last page. While I do not see that happening with List, I prefer to leave this freedom to engineering.
Therefore, the only indication of whether we've reached the end is whether the result has any pagination token.

pinecone/grpc/index_grpc.py
```diff
@@ -428,6 +433,55 @@ def update(
...
+            try:
+                results = self.list_paginated(**kwargs)
+            except Exception as e:
+                raise e
+            
+            if len(results.vectors) > 0:
+                yield [v.id for v in results.vectors]
+            
+            full_page = len(results.vectors) == limit
+            if results.pagination and results.pagination.next and full_page:
```
opbenesh, line 480: Same as above: it's valid for a page to contain less than `limit` results without being the last page. While I do not see that happening with List, I prefer to leave this freedom to engineering.
Therefore, the only indication of whether we've reached the end is whether the result has any pagination token.

opbenesh (changes requested):

pinecone/data/index.py
jhamon: What you're describing is the current behavior, but I don't think it's valid. I reported it as a bug and your team filed an Asana issue for it. I added this "full page" check so that the behavior would be deterministic in testing and not fire off additional calls to fetch an empty array.

opbenesh: The bug was specifically about adding an extra empty page at the end of the pagination loop, but the generic case of sending <limit results is valid - see the [API Pagination PRD](https://docs.google.com/document/d/19lJBFhJ2VqBSGwHzenUk40Oe64mPf3d0QttY2PP-tiQ/edit#heading=h.hgq4984r4biv):
> Users should not assume that a page containing fewer results than the limit value is necessarily the last one - and instead should always check the pagination.next field.

austin-denoble (approved):
The code here looks good to me, helpful to go through this right before implementing on TypeScript.

I tried pulling this branch down and testing locally and I've run into issues with both `list` and `list_paginated`, although it seems to be the same error:

```
>>> pc.Index('step-test').list_paginated(namespace="test-list")
curl -X GET 'https://step-test-e6dddb2.svc.us-east-1-aws.pinecone.io/vectors/list?namespace=test-list' -H 'Accept: application/json' -H 'User-Agent: python-client-3.0.2 (urllib3:2.0.7)' -H 'Api-Key: redacted'

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/austin/workspace/pinecone-python-client/pinecone/utils/error_handling.py", line 10, in inner_func
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/data/index.py", line 523, in list_paginated
    return self._vector_api.list(**args_dict, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/core/client/api_client.py", line 772, in __call__
    return self.callable(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/core/client/api/vector_operations_api.py", line 712, in __list
    return self.call_with_http_info(**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/core/client/api_client.py", line 834, in call_with_http_info
    return self.api_client.call_api(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/core/client/api_client.py", line 409, in call_api
    return self.__call_api(resource_path, method,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/core/client/api_client.py", line 224, in __call_api
    return_data = self.deserialize(
                  ^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/core/client/api_client.py", line 325, in deserialize
    deserialized_data = validate_and_convert_types(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/core/client/model_utils.py", line 1539, in validate_and_convert_types
    converted_instance = attempt_convert_item(
                         ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/austin/workspace/pinecone-python-client/pinecone/core/client/model_utils.py", line 1421, in attempt_convert_item
    raise get_type_error(input_value, path_to_item, valid_classes,
pinecone.core.client.exceptions.PineconeApiTypeError: Invalid type for variable 'received_data'. Required value type is ListResponse and passed type was str at ['received_data']
```

Your integration tests seem to be passing though so I'm assuming this is something on my end. I'll keep playing around with it.

pinecone/data/index.py
tomer-w: I agree with Ben here. In the generic case we should have the ability to shards result and than it is common practice to return less than the full page and still have continuation token.

pinecone/data/index.py
```diff
@@ -502,6 +502,40 @@ def describe_index_stats(
...
+                ("limit", limit),
+                ("namespace", namespace),
+                ("pagination_token", pagination_token),
+            ]
+        )
+        return self._vector_api.list(**args_dict, **kwargs)
+
+    @validate_and_convert_errors
+    def list(self, **kwargs):
+        limit = kwargs.get("limit", 100)
```
austin-denoble, line 527: When we've settled on what the functions are going to look like we should add some docstrings here and in `index_grpc.py`.

tests/integration/data/test_list.py
```diff
@@ -0,0 +1,100 @@
...
+        assert len(results.vectors) == 0
+        # assert results.pagination == None
+
+    def test_list_no_args(self, idx):
+        results = idx.list_paginated()
+        
+        assert results != None
+        assert len(results.vectors) == 9
+        assert results.namespace == ''
+        # assert results.pagination == None
```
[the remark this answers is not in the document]
jhamon: Yeah, still need these. I filed a bug with the engine team to address this issue.

jhamon:
I removed the full_page check and added in docstrings.

PR #310, opened 2024-02-13
upsert_from_dataframe: Hide all progressbars if !show_progress

daverigby:
## Problem

For GRPCIndex.upsert_from_dataframe() there is a second progressbar
used when use_async_requests is true (the default) - 'collecting async
responses'. However this progressbar is always shown, even if
show_progress is false.

## Solution

Make it respect the show_progress argument.

## Type of Change


jhamon (approved):
LGTM thanks

PR #312, opened 2024-02-14
Adjust supported python versions

jhamon:
## Problem

## Solution

## Type of Change

## Test Plan


.github/workflows/testing-dependency.yaml
```diff
@@ -96,8 +91,58 @@ jobs:
...
+    strategy:
+      max-parallel: 5
+      fail-fast: false
+      matrix:
+        python-version:
+          - 3.11
+          - 3.12
+        grpcio-version:
+          - 1.59.3
+          - 1.60.0
```
jhamon, line 107: This range is different for 3.11 and 3.12 vs earlier versions.

.github/workflows/testing-dependency.yaml
```diff
@@ -96,8 +91,58 @@ jobs:
...
+          - 3.1.3
+          - 4.3.3
+        protobuf-version:
+          - 3.20.3
+        googleapis-common-protos-version:
+          - 1.53.0
+          - 1.62.0
+        grpc-gateway-protoc-gen-openapiv2-version:
+          - 0.1.0
+    steps:
```
jhamon, line 118: These test steps are the same as above, but unfortunately github actions doesn't support yaml anchors that would enable easy reuse in file https://github.com/actions/runner/issues/1182

pyproject.toml
```diff
@@ -47,15 +48,21 @@ classifiers=[
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.13"
+python = "^3.8"
```
jhamon, line 51: This was the requested change that motivated me to start poking around with this.

pyproject.toml
```diff
@@ -82,11 +89,11 @@ numpy = [
 ]
 pandas = ">=1.3.5"
 pdoc = "^14.1.0"
-pytest = "7.2.0"
+pytest = "8.0.0"
 pytest-asyncio = "0.15.1"
 pytest-cov = "2.10.1"
 pytest-mock = "3.6.1"
-pytest-timeout = ">=1.4.2"
+pytest-timeout = "2.2.0"
```
jhamon, line 96: We needed to bump these dev dependencies to work with python 3.12.

PR #313, opened 2024-02-16
Change regex validation for names

fsxfreak:
## Problem

We allow index names to be alphanumeric with dashes in between, and require the first and last character be alphanumeric. The existing regex doesn't allow for that, so here's the proposed regex 

## Solution

Cross checked with internal specs

## Type of Change

## Test Plan

Played around with the regex in https://regex101.com/

### DIFF
--- a/pinecone/core/client/model/create_collection_request.py
+++ b/pinecone/core/client/model/create_collection_request.py
@@ -63,7 +63,7 @@ class CreateCollectionRequest(ModelNormal):
             'max_length': 45,
             'min_length': 1,
             'regex': {
-                'pattern': r'^[A-Za-z0-9].*[A-Za-z0-9]$',  # noqa: E501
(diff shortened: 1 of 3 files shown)


jhamon (commented):
Hey, thanks for helping out again here. The files you're attempting to edit (and everything under `pinecone/core/`) are generated from our privately held openapi spec. So the way to update these values is to make a change to the spec and then regenerate this part of the client. If you ask in #apis slack channel people can walk you through these steps.

fsxfreak:
> Hey, thanks for helping out again here. The files you're attempting to edit (and everything under `pinecone/core/`) are generated from our privately held openapi spec. So the way to update these values is to make a change to the spec and then regenerate this part of the client. If you ask in #apis slack channel people can walk you through these steps.

thanks, let me open another PR after I figure out where the source of truth spec is

PR #314, opened 2024-02-18
Update pytest-timeout to support python >= 3.12

mjvankampen:
## Problem

I could not run unit tests because pytest-timeout relies on distutils which was removed in python 3.12

## Solution

Updated pytest-timeout

## Type of Change

## Test Plan

Run all unit tests. Does not impact the actual client.


jhamon:
Looks good, thanks.

PR #315, opened 2024-02-21
Sync models from pinecone-protos

fsxfreak:
## Problem

See context in #313. This PR also includes other changes such as renaming vector_operations_api to data_plane_api and updating documentation.

## Solution

Followed steps to regenerate the python client from the protos.

## Type of Change

### DIFF
--- a/pinecone/core/client/model/create_collection_request.py
+++ b/pinecone/core/client/model/create_collection_request.py
@@ -62,9 +62,6 @@ class CreateCollectionRequest(ModelNormal):
         ('name',): {
             'max_length': 45,
             'min_length': 1,
-            'regex': {
-                'pattern': r'^[A-Za-z0-9].*[A-Za-z0-9]$',  # noqa: E501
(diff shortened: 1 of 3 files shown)


jhamon (changes requested):
~~I'm a bit puzzled why the tests did not kick off for this PR. But I would expect a few things to fail because of `ManageIndexesApi` getting renamed to `DataPlaneApi`. This class is referenced inside of `pinecone/control/pinecone.py` and `pinecone/control/index_host_store.py` which need to be changed to account for the rename.~~

I misread the diff because of how github presented it, although some stuff did get renamed that should require changes outside the generated directory. E.g. `VectorServerApi` to `DataPlaneApi` and `delete` to `delete_alt`. I pushed a whitespace change to trigger the tests which should hopefully highlight that change is needed; I'm not sure why they didn't run when you opened this PR.

jhamon (approved):

PR #316, opened 2024-02-23
Fix minor README docs issues in client reference

austin-denoble:
## Problem
@haruska pointed out that the headers in the SDK client reference navigation bar were a little off for "Create Index" showing "Create a pod index" only. The formatting also felt a bit off with everything being aligned.

https://sdk.pinecone.io/python/pinecone.html

While fixing this I also noticed the link for the `CONTRIBUTING.md` file is relative in the README, which doesn't work when linking from our generated docs.

## Solution
- Promote "Indexes" to a top-level section to visually break up the nav list a bit, and then swap `Create Index` to the header and moving `Create a serverless index` and `Create a pod index` to sub-headers of that.
- Link directly to github for the `CONTRIBUTING.md` file. 

## Type of Change

## Test Plan

|Before|After|
| ------------- | ------------- |
|Screenshot 2024-02-23 at 12 55 08 PM|Screenshot 2024-02-23 at 12 55 32 PM|

You can pull this branch down and build the docs locally. You'll need to make sure to install the grpc extras with poetry:

```
poetry install --extras "grpc"
poetry run pdoc pinecone '!pinecone.core' '!pinecone.utils' --favicon ./favicon-32x32.png --docformat google -o ./docs
```

This will build the docs in the `/docs` folder under the project folder. You can open the `index.html` file locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 # Pinecone Python Client &middot; ![License](https://img.shields.io/github/license/pinecone-io/pinecone-python-client?color=orange) [![CI](https://github.com/pinecone-io/pinecone-python-client/actions/workflows/merge.yaml/badge.svg)](https://github.com/pinecone-io/pinecone-python-client/actions/workflows/merge.yaml)
 
-
 The official Pinecone Python client.
 
 For more information, see the docs at https://www.pinecone.io/docs/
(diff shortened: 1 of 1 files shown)


haruska (approved):
Thanks for cleaning this up!

jhamon (approved):

PR #319, opened 2024-03-07
[SDK build process] Implement `make regenerate` command

jhamon:
## Problem

We're updating our SDK build process to consume openapi/grpc artifacts published into a private gcloud bucket.

## Solution

- Move openapi python templates from private `pinecone-protos` repository into the `codegen/` directory (reviewers: since these files are just being copied over from elsewhere without modification, they don't need close reviewing)
- Add Makefile build commands to regenerate the openapi rest core and install new grpc stubs
- Implemented Makefile gcloud CLI commands using a dockerized version of the CLI.
- Add build-related directories `downloads/` and `gen/` to `.gitignore` so they are not checked in.

## Usage

Run these make commands to update the generated parts of the SDK:

```
make gcloud-auth
make regenerate
```

## Type of Change

## Test Plan

I have another branch where I've tried running these commands and then executed tests with the updated client. I wanted to keep the size of this diff down to a manageable size so I did not commit those changes. See here https://github.com/pinecone-io/pinecone-python-client/pull/320


Makefile
```diff
@@ -1,48 +1,48 @@
...
+	docker run --rm \
+		--volumes-from gcloud-config \
+		-v ${mkfile_path}:/workspace \
+		-w /workspace \
+		gcr.io/google.com/cloudsdktool/google-cloud-cli \
+		gcloud storage cp gs://api-codegen/_latest downloads --recursive
+
+regenerate: clean download
+	# Remove problematic query param
+	jq 'walk(if type == "array" then map(if type == "object" then select(.name != "filter") else . end) else . end)' downloads/_latest/openapi/pinecone_api.json > tmp.json
```
jhamon, line 31: See https://github.com/OpenAPITools/openapi-generator/issues/1587

jhamon, line 31: Query params with type `object` seem to break the python generator. See https://github.com/OpenAPITools/openapi-generator/issues/1587

Makefile
```diff
@@ -1,48 +1,48 @@
...
+	mv tmp.json downloads/_latest/openapi/pinecone_api.json
```
jhamon, line 32: Query params with type `object` seem to break the python generator. See https://github.com/OpenAPITools/openapi-generator/issues/1587

haruska (changes requested):
Related to https://github.com/pinecone-io/apis/pull/50#issuecomment-1986085746

PR #321, opened 2024-03-12
[Fix] Configuring SSL proxy via openapi_config object

jhamon:
## Problem

A few different problems being solved here:
- `Pinecone` class accepts an optional param, `openapi_config`. When this is passed (usually as a vehicle for SSL configurations), it currently clobbers the `api_key` param so the user sees an error message about not providing an api_key (even though they did pass it) if they attempt to perform a control plane operation

```python
from pinecone import Pinecone
from pinecone.core.client.configuration import Configuration as OpenApiConfiguration

openapi_config = OpenApiConfiguration()
openapi_config.ssl_ca_cert = '/path/to/cert'

pc = Pinecone(api_key='key, openapi_config=openapi_config)
pc.list_indexes() // error: No api-key provided
```

- In a related issue, the `openapi_config` (with SSL configs) was not being correctly passed through to the underlying `DataPlaneApi` for data calls. So users with custom network configurations requiring SSL config would see SSL validation failures when attempting data operations.

```python
from pinecone import Pinecone
from pinecone.core.client.configuration import Configuration as OpenApiConfiguration

openapi_config = OpenApiConfiguration()
openapi_config.ssl_ca_cert = '/path/to/cert'

pc = Pinecone(api_key='key, openapi_config=openapi_config)
pc.list_indexes() // error: No api-key provided
```

## Solution

- Adjust the ConfigBuilder to avoid clobbering API key
- Move some logic into a util function so that behavior will be consistent across both data and control planes
- Ensure configuration is passed to from the `Pinecone` object to the index client
- deepcopy the openapi_config object before modifying it so that index-specific host changes do clobber control plane or calls to other indexes.

## Future work

- In the future, we should deprecate `openapi_config` and have some way of passing SSL config without all the baggage that comes with this OpenApiConfiguration object. This config object is an undocumented holdover from earlier versions of the client and breaks the abstraction the client is trying to provide to smooth out the UX of the generated SDK code.

## Type of Change

## Test Plan

- Added tests

### DIFF
--- a/pinecone/config/config.py
+++ b/pinecone/config/config.py
@@ -46,10 +46,10 @@ class ConfigBuilder:
         if not host:
             raise PineconeConfigurationError("You haven't specified a host.")
 
-        openapi_config = (
-            openapi_config
-            or kwargs.pop("openapi_config", None)
-            or OpenApiConfigFactory.build(api_key=api_key, host=host)
-        )
+        if openapi_config:
+            openapi_config.host = host
+            openapi_config.api_key = {"ApiKeyAuth": api_key}
+        else:
+            openapi_config = OpenApiConfigFactory.build(api_key=api_key, host=host)
 
         return Config(api_key, host, openapi_config, additional_headers)
--- a/pinecone/config/config.py
+++ b/pinecone/config/config.py
@@ -1,5 +1,6 @@
 from typing import NamedTuple, Optional, Dict
 import os
+import copy
 
 from pinecone.exceptions import PineconeConfigurationError
 from pinecone.config.openapi import OpenApiConfigFactory
@@ -47,6 +48,7 @@ class ConfigBuilder:
             raise PineconeConfigurationError("You haven't specified a host.")
 
         if openapi_config:
+            openapi_config = copy.deepcopy(openapi_config)
             openapi_config.host = host
             openapi_config.api_key = {"ApiKeyAuth": api_key}
         else:
--- a/pinecone/control/pinecone.py
+++ b/pinecone/control/pinecone.py
@@ -515,12 +515,26 @@ class Pinecone:
             raise ValueError("Either name or host must be specified")
         
         pt = kwargs.pop('pool_threads', None) or self.pool_threads
+        api_key = self.config.api_key
+        openapi_config = self.config.openapi_config
 
         if host != '':
             # Use host url if it is provided
-            return Index(api_key=self.config.api_key, host=normalize_host(host), pool_threads=pt, **kwargs)
+            return Index(
+                host=normalize_host(host),
+                api_key=api_key,
+                pool_threads=pt,
+                openapi_config=openapi_config,
+                **kwargs
+            )
 
         if name != '':
             # Otherwise, get host url from describe_index using the index name
             index_host = self.index_host_store.get_host(self.index_api, self.config, name)
-            return Index(api_key=self.config.api_key, host=index_host, pool_threads=pt, **kwargs)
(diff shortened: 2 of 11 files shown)


pinecone/config/config.py
```diff
@@ -46,10 +46,10 @@ def build(
         if not host:
             raise PineconeConfigurationError("You haven't specified a host.")
 
-        openapi_config = (
-            openapi_config
-            or kwargs.pop("openapi_config", None)
```
jhamon, line 51: `or kwargs.pop("openapi_config", None)` wasn't doing anything, since `openapi_config` is a named param and the key should never appear in `kwargs`.

pinecone/config/config.py
```diff
@@ -46,10 +46,10 @@ def build(
...
-            or OpenApiConfigFactory.build(api_key=api_key, host=host)
-        )
+        if openapi_config:
+            openapi_config.host = host
+            openapi_config.api_key = {"ApiKeyAuth": api_key}
```
jhamon, line 51: When the user provides this object with some configuration in it, we want to merge the api key and host settings into it rather than building a fresh object.

pinecone/control/pinecone.py
```diff
@@ -85,25 +84,20 @@ def __init__(
...
         your API key** which forms part of a required authentication header. Default: `false`
         """
-        if config or kwargs.get("config"):
-            configKwarg = config or kwargs.get("config")
-            if not isinstance(configKwarg, Config):
+        if config:
+            if not isinstance(config, Config):
                 raise TypeError("config must be of type pinecone.config.Config")
             else:
-                self.config = configKwarg
```
jhamon, line 93: This was more cleanup since `config` is a named param that should never appear in `kwargs`

pinecone/control/pinecone.py
```diff
@@ -521,12 +515,26 @@ def Index(self, name: str = '', host: str = '', **kwargs):
...
             # Otherwise, get host url from describe_index using the index name
             index_host = self.index_host_store.get_host(self.index_api, self.config, name)
-            return Index(api_key=self.config.api_key, host=index_host, pool_threads=pt, **kwargs)
+            return Index(
+                host=index_host,
+                api_key=api_key,
+                pool_threads=pt,
+                openapi_config=openapi_config,
+                **kwargs
+            )
```
ssmith-pc, line 540: This may be out of the scope for this change since you're just adding the `openapi_config` param, but it looks like these two blocks are largely identical, differing only by how the `host` param is init'd

pinecone/utils/setup_openapi_client.py
```diff
@@ -0,0 +1,14 @@
+from pinecone.core.client.api_client import ApiClient
+from .user_agent import get_user_agent
+
+def setup_openapi_client(api_klass, config, pool_threads):
```
austin-denoble, line 4: nit: `_klass`?

tests/integration/data/test_openapi_configuration.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+import os
+
+from pinecone import Pinecone
+from pinecone.core.client.configuration import Configuration as OpenApiConfiguration
+from urllib3 import make_headers
+
+@pytest.mark.skipif(os.getenv('USE_GRPC') != 'false', reason='Only test when using REST')
```
austin-denoble, line 8: Nice. 👍

austin-denoble (approved):
Nice job tracking this one down and cleaning that up. 🚢

PR #324, opened 2024-03-14
Allow clients to tag requests with a source_tag

ssmith-pc:
## Problem
Need to allow clients to include a `source_tag` to identify the source of their requests.

## Solution
Allow clients to specify a `source_tag` field in the client constructor, that will be used to identify the traffic source, if applicable. 

Example:
```python
from pinecone import Pinecone

pc = Pinecone(api_key='foo', source_tag='bar')

pc.list_indexes()
```

This would cause the user-agent to get a value like:
```
User-Agent: 'python-client-3.1.0 (urllib3:2.0.7); source_tag=bar'
```

## Type of Change

## Testing

- [ ] Tests are passing
- [ ] Verify source_tag included in user-agent in control plane and data plane (REST and gRPC)


pinecone/config/config.py
```diff
@@ -41,6 +42,7 @@ def build(
         api_key = api_key or kwargs.pop("api_key", None) or os.getenv("PINECONE_API_KEY")
         host = host or kwargs.pop("host", None)
         host = normalize_host(host)
+        source_partner = kwargs.pop("source_partner", None)
```
ssmith-pc, line 45: Currently I'm thinking this would go in `kwargs` vs a named parameter as this would be more of an internal field we'd just ask partners to include in their integrations.

jhamon: I agree `kwargs` makes sense here because we don't want to invite randos to set strange values. Perhaps partner/integration name and version should be passed separately so that we can have greater control over formatting.  For your example, perhaps I would prefer `pc = Pinecone(api_key='foo', source_id='bar', source_version='1.2')`. That way you don't end up with one person passing `foo/1.2` and another passing `bar 3.5` and causing headaches in parsing when it's time to try to turn this all into some kind of dashboard.

jhamon:
Something else to keep in mind is we would like to track usage of SDKs coming from our own higher-level tools in addition to external integration partners. For example, 

- [canopy](https://github.com/pinecone-io/canopy) depends on the python sdk
- [spark-connector](https://github.com/pinecone-io/spark-pinecone/tree/main) depends on java sdk

This doesn't change a ton about the general strategy, but it makes me think having `partner` in the name of these params isn't quite right.

jhamon:
We should also reevaluate some of the cruft that's currently being passed in the user-agent strings. I've been here for a year and never had a need to know anything about what urllib3 version was in use.

pinecone/config/config.py
ssmith-pc: Agree on separating the name from version.

ssmith-pc:
> We should also reevaluate some of the cruft that's currently being passed in the user-agent strings. I've been here for a year and never had a need to know anything about what urllib3 version was in use.

That's good to know that there isn't much relying on what's there currently. Would be nice to identify the high value information we'd like to capture and include. 

I can understand how it *could* be helpful to know what http client/version is being used to access the service, particularly if there is a known bug with a particular version that causes trouble for our service, as this would likely help CS triage and help.

ssmith-pc:
> This doesn't change a ton about the general strategy, but it makes me think having `partner` in the name of these params isn't quite right.

Agree we should capture those clients, but I do wonder if we should try to keep the partner info separate. A lot of that depends on what we end up doing in the back-end, like how this info is extracted and matched up to what's in our partner CRM, etc. 

In addition to setting the `source_*` fields, we could include a separate flag in *our* clients that identifies it as an internally owned integration.

pinecone/config/config.py
ssmith-pc: This is done in the latest version

jhamon:
FYI there is a minimal unit test file called `test_user_agent.py` that you may want to expand with a few more test cases to make sure the outputted string looks reasonable across a handful of different normalization cases.

ssmith-pc:
> FYI there is a minimal unit test file called `tests/unit/utils/test_user_agent.py` that you may want to expand with a few more test cases to make sure the resulting string looks reasonable across a handful of different normalization cases.

I was working on that concurrently with your comment -- just pushed

jhamon (approved):
Approach and tests look good to me.

It can be tackled in a different PR if you want, but I think you'll need to go delving into GRPC `base.py` code and look at how `additional_metadata` gets passed. Unfortunately, I think those data calls are not sending the same user agent currently as gets sent with REST calls.

PR #325, opened 2024-03-19
Revise proxy configuration, add integration testing

jhamon:
## Problem

We want to expose proxy configuration fields as top-level config params and move away from users passing an `OpenApiConfiguration` instance with `openapi_config` (as was done in 2.x versions of the sdk). Passing these objects is not something we documented or had shown as a named configuration param, but is something people assumed would work as a hangover from the way things were done in the old client.

Currently OpenApi generated code is part of our implementation, but we may want to swap out something else in the future so it doesn't make sense to use this `OpenApiConfiguration` object as part of our public interface. And since the openapi config object also needs to be configured with api key and host, it competes with these other config params in a way that is pretty confusing.

## Solution

- Add new configuration parameters for `proxy_url`, `proxy_headers`, `ssl_ca_certs`, and `ssl_verify`.
- Document the use of these properties in the README and doc strings
- Emit a warning message when users are passing `openapi_config` encouraging them to use the new properties.
- Implement integration tests that run with [mitmproxy](https://mitmproxy.org/) docker image. For now these are only run in local development.

## Usage

```python
from pinecone import Pinecone
import urllib3 import make_headers

pc = Pinecone(
    api_key="YOUR_API_KEY",
    proxy_url='https://your-proxy.com',
    proxy_headers=make_headers(proxy_basic_auth='username:password'),
    ssl_ca_certs='path/to/cert-bundle.pem'
)
```

This will work, but should emit a deprecation warning:

```python
from pinecone import Pinecone
from pinecone.core.client.configuration import OpenApiConfiguration
import urllib3 import make_headers

config = OpenApiConfiguration()
config.ssl_ca_certs = 'path/to/cert-bundle.pem'
config.proxy_headers = make_headers(proxy_basic_auth='username:password')
config.proxy_url = 'https://your-proxy.com'

pc = Pinecone(
    api_key="YOUR_API_KEY",
    openapi_config=config
) # emits deprecation notice

```

## Future work
We will need to figure out how to deploy the mitmproxy elsewhere (cloud run?) before running tests in CI because Github Actions seems to have a lot of issues with running these docker containers and debugging the errors has proven quite difficult. While GHA can run docker containers, mounting a volume with the required certs causes the container to crash. I'm guessing this is some kind of file permissions issue, but don't have much visibility for debugging.

## Type of Change

## Test Plan

Run new tests locally with `PINECONE_API_KEY='foo' PINECONE_INDEX_NAME='bar' poetry run pytest tests/integration/proxy_config/ -s -v`


.github/workflows/testing-dependency.yaml
```diff
@@ -158,3 +158,19 @@ jobs:
           index_name: '${{ needs.dependency-matrix-setup.outputs.index_name }}'
           PINECONE_API_KEY: '${{ secrets.PINECONE_API_KEY }}'
           urllib3_version: '${{ matrix.urllib3_version }}'
+
+  deps-cleanup:
```
jhamon, line 162: This is not really related to the config changes, but keeps us from getting values due to 20 index per-project limit.

.github/workflows/testing-integration.yaml
```diff
@@ -3,6 +3,54 @@ name: "Integration Tests"
   workflow_call: {}
 
 jobs:
+  # setup-index:
```
jhamon, line 6: Leaving this commented code in as I have plans to bring back CI integration tests in a future PR.

tests/integration/proxy_config/.mitm/proxy1/mitmproxy-ca-cert.cer
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
jhamon, line 1: All of these cert / key things are generated by mitmproxy.

pinecone/config/config.py
```diff
@@ -47,11 +52,28 @@ def build(
...
+            openapi_config = OpenApiConfigFactory.copy(openapi_config=openapi_config, api_key=config.api_key, host=config.host)
+        elif openapi_config is None:
+            openapi_config = OpenApiConfigFactory.build(api_key=config.api_key, host=config.host)
 
-        return Config(api_key, host, openapi_config, additional_headers)
\ No newline at end of file
+        # Check if value passed before overriding any values present
+        # in the openapi_config. This means if the user has passed 
+        # an openapi_config object and a kwarg for the same setting,
+        # the kwarg will take precedence.
```
austin-denoble, line 69: I may be reading this incorrectly, but it doesn't look like we're doing anything with `kwargs` here or in `build`?

Is the comment meant to be referencing overriding `openapi_config` with `config`?

tests/integration/proxy_config/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+    },
+    'proxy2': {
+        'name': 'proxy2',
+        'port': 8081, 
+        'ssl_ca_certs': os.path.abspath('./tests/integration/proxy_config/.mitm/proxy2'),
+        'auth': ('testuser', 'testpassword')
+    }
+}
+
+def docker_command(proxy):
```
austin-denoble, line 22: Nice work getting all of this set up, thank you.

austin-denoble (approved):
Thanks for putting so much effort into this, I think what you've learned working on the implementation here will help us across the board. Overall this all looks good to me, thanks for adding test coverage generally. 🚢

pinecone/config/config.py
jhamon: The `config` object is built using kwarg values. It happens in a different spot, but that's where those values come from. Sorry it's not more clear. Things get confusing when trying to manage and merge these different sources of configuration, which is a big reason I want to deprecate passing `openapi_config`.

PR #326, opened 2024-03-22
Add grpc unit test run, expand testing of VectorFactory

jhamon:
## Problem

There is a small number of unit tests that should only get run when grpc dependencies are installed. These were previously omitted from CI by mistake.

## Solution

Add a step to run these grpc steps. Make it conditional on the `use_grpc` test matrix param.

## Type of Change

### DIFF
--- a/.github/workflows/testing-unit.yaml
+++ b/.github/workflows/testing-unit.yaml
@@ -31,7 +31,7 @@ jobs:
       - name: Run unit tests (REST)
         run: poetry run pytest --cov=pinecone --timeout=120 tests/unit
       - name: Run unit tests (GRPC)
-        if: ${{ matrix.use_grpc == 'true' }}
+        if: ${{ matrix.use_grpc == true }}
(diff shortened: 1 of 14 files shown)


austin-denoble (approved):
Nice, thanks! 🚢

ssmith-pc:
Thanks!

.github/workflows/testing-unit.yaml
```diff
@@ -28,8 +28,11 @@ jobs:
         with:
           include_grpc: '${{ matrix.use_grpc }}'
           include_types: true
-      - name: Run unit tests
+      - name: Run unit tests (REST)
```
ssmith-pc, line 31: I think these do cover the REST surface area, but also a lot of the "other" stuff like common utils, etc.

ssmith-pc (approved):

.github/workflows/testing-unit.yaml
jhamon: Yeah. They're really meant to be "everything that doesn't require GRPC dependencies" unit tests. But that doesn't roll off the tongue 🤣

PR #333, opened 2024-04-19
pinecone-client > pinecone

tdonia:
Per https://github.com/pypi/support/issues/2432#issuecomment-2066632560 -- we have the `pinecone` package name on PyPI

## Problem

This is a rough first pass at updating the package name on PyPI to `pinecone`. It doesn't attempt preserve the prior package workflow -- we may want to continue to publish to both for some period of time? We may also want to update the publishing workflow first, then doc changes later.

## Solution

Updated Readmes, inline docs, and github workflows to support `pinecone` instead of `pinecone-client`

## Type of Change

- [X] This change requires a documentation update
- [X] None of the above: (explain here): Package name update

## Test Plan

- Does it publish to the new namespace?
- How much work do we have to do to help users migrate from `pinecone-client` -- i.e. are there any automatic redirects, or is this a manual deprecation?

### DIFF
--- a/.github/actions/create-index-legacy/action.yml
+++ b/.github/actions/create-index-legacy/action.yml
@@ -32,10 +32,10 @@ runs:
       with:
         python-version: 3.9
 
-    - name: Install pinecone-client
+    - name: Install pinecone
       shell: bash
       run: |
-        pip install pinecone-client==${{ inputs.pinecone_client_version }}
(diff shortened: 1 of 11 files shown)


aulorbe (approved):
Just lurking, but LGTM!

ssmith-pc:
I don't have deep experience with PyPI, but this seems relevant: 
https://peps.python.org/pep-0423/#how-to-rename-a-project

PR #334, opened 2024-04-22
Improve upsert throughput by 3x

daverigby:
## Problem

Python SDK upsert throughput is low compared to other SDKs - for example I can achive 880 vector upserts/sec with the Python SDK, compared to 3500 upserts/sec with the Java SDK. 

Profiling the Python SDK performing these upserts shows a large  percentage of time in gRPC / protobuf serialisation / deserialisation.

## Solution

Upgrade protobuf from v3 to v4. This adds a number of performance improvements in parsing / serialization as documented at https://protobuf.dev/news/2022-05-06/#python-updates

This increases upsert() throughput by 3x (measured by upserting 1M 768 dimension indexes to a pod-based index in batches of 500): 

* Before:  880 vectors/sec
* After:  2580 vectors/sec

As per the documentation, this results in an incompatible change with the _generated_ Python code, so this depends on a related change to pinecone-protos to change the version of protobuf used to generate the Python code there.

## Type of Change

- [x] None of the above: Performance improvement.

## Test Plan

Use existing regression tests.


pyproject.toml
```diff
@@ -63,10 +63,10 @@ grpcio = [
   { version = ">=1.44.0", optional = true, python = "^3.8,<3.11" },
   { version = ">=1.59.0", optional = true, python = "^3.11" }
 ]
-grpc-gateway-protoc-gen-openapiv2 = { version = "0.1.0", optional = true }
 googleapis-common-protos = { version = ">=1.53.0", optional = true }
 lz4 = { version = ">=3.1.3", optional = true }
-protobuf = { version = "~=3.20.0", optional = true }
+protobuf = { version = "^4.25", optional = true }
+protoc-gen-openapiv2 = {version = "^0.0.1", optional = true }
```
jhamon, line 69: This `protoc-gen-openapiv2` dependency looks a bit sus having only 1 release, the [underlying repo](https://github.com/unionai-oss/protoc-gen-openapiv2) having no stars, and otherwise looking unofficial. How did you come across this package?

Also, is this actually something the user's app depends on at runtime or is it only used during code generation? So essentially a dev-dependency. If it's only used in the codegen process, we don't need to declare it in this deps list which are for the consumers of this package.

jhamon (approved):
Thanks for pushing this forward. A couple things:
- I left a question below about the new dependency and if it's needed by the user at runtime 
- Also, I want to make sure I understand the process you followed to generate this so we don't blow away your changes next time we make an update. We can discuss this over slack.

Except for that, I think this should be ready to merge. I appreciate that you added the nice error message.

pyproject.toml
daverigby: It's pretty much as sus as the package it replaced (grpc-gateway-protoc-gen-openapiv2) ;)

I came across it when searching for a newer package providing protobuf v4 Python generation for google.api.annotations - In essence this is two different people fixing a problem in upstream grpc-gateway where they fail to automatically generate code for dependant .proto files for Python - see https://github.com/grpc-ecosystem/grpc-gateway/issues/1785

The alternative would be to essentially roll what this package does (generate Python bindings from `google.api.annotations` etc) into our own package.

austin-denoble: Regarding @jhamon's question around dev dependency vs. runtime dependency: do we need to move this at all or is it essentially replacing what was there and is still needed for runtime?

austin-denoble (approved):
Thanks for tracking this one down and adding in some additional improvements. Also pushing up the corresponding changes to the `pinecone-protos` repo, thanks!

pyproject.toml
daverigby: It's replacing the previous package, so we still need it at runtime (see the import on line 18 of pinecone/core/grpc/protos/vector_service_pb2.py):

```python
from protoc_gen_openapiv2.options import annotations_pb2 as protoc__gen__openapiv2_dot_options_dot_annotations__pb2
```

PR #337, opened 2024-05-03
Feature-based architecture exploration

jhamon:
## Problem

## Solution

## Type of Change

## Test Plan


scripts/feature-gen.py
```diff
@@ -0,0 +1,249 @@
...
+        else:
+            validations.append((f'if {propName} is not None and {propName} not in {propConfig["enum"]}', f'raise PineconeApiValueError("{propName} must be one of {propConfig["enum"]}")'))
+
+type_translation = {
+    'string': 'str',
+    'integer': 'int',
+    'boolean': 'bool',
+    'array': 'List',
+    'object': 'Dict',
+    'number': 'float'
```
aulorbe, line 63: OOC: why do you do `List` (instead of `list`) and `Dict` instead of (`dict`) here?

aulorbe: (Mostly ask b/c I think [Python's moving towards](https://docs.python.org/3/library/typing.html#typing.List) deprecating the `typing.<whatever>` types in favor of built-ins)

jhamon:
I learned a lot from scribbling some ideas here as a POC, but I think it makes sense to pursue code generation in a more careful and rigorous way somewhere outside of this repository.

PR #342, opened 2024-05-13
Add better error messages for mistaken `from_texts` and `from_documents`

jhamon:
## Problem

Sometimes people using Pinecone with Langchain accidentally try to invoke `from_texts` and `from_documents` methods on our `Pinecone` class even though these are not methods by us. This is mostly name-related confusion because Langchain used to have an export called `Pinecone` that later got renamed to the less ambiguous `PineconeVectorStore`.

## Solution

Add some stub methods to the `Pinecone` class to guide users in the right direction.

Screenshot 2024-05-13 at 2 36 51 PM

## Type of Change

## Test Plan

Added unit tests. And also an integration test to verify the docs link resolves to a valid page.

### DIFF
--- a/pinecone/utils/docslinks.py
+++ b/pinecone/utils/docslinks.py
@@ -1,0 +1,3 @@
+docslinks = {
+    'LANGCHAIN_IMPORT_KB_ARTICLE': "https://docs.pinecone.io/troubleshooting/pinecone-attribute-errors-with-langchain"
+}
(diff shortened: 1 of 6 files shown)


junefish (approved):
Thanks!

pinecone/utils/docslinks.py
```diff
@@ -0,0 +1,4 @@
+docslinks = {
```
austin-denoble, line 1: Nice, good thinking. 👍

austin-denoble (approved):
Looks good

PR #349, opened 2024-05-22
Make GRPC a drop in replacement

mcpaddy:
## Problem

To make the GRPC version a drop-in for the normal client. For example LangChain.

## Solution

Alias the function get to result. 

## Type of Change

## Test Plan

Test using LangChain and my own code. There should be no need for this with the change.

```
    if mode  == 'GRPC':
      [async_result.result() for async_result in async_results]
    else:
      [async_result.get() for async_result in async_results]
```


jhamon:
I'm addressing this in another PR in a way that will include tests, etc.

PR #350, opened 2024-06-04
Remove unused constants

jhamon:
## Problem

I noticed there are a few constants defined that are not currently being used anywhere.

## Solution

- Search in codebase to confirm these are no longer in use.
- Remove them.

## Type of Change

- [x] Chore (cleanup dead code)

## Test Plan

See tests still pass.


austin-denoble (approved):
🚢

PR #351, opened 2024-06-04
Allow colon inside source tags

jhamon:
## Problem

Some partners have requested that colon, `:`, be added to the set of allowed characters inside source tags.

## Solution

Add colon to the set of allowed characters.

## Type of Change

## Test Plan

Unit tests updated to reflect new allowed character.

### DIFF
--- a/pinecone/utils/user_agent.py
+++ b/pinecone/utils/user_agent.py
@@ -11,7 +11,7 @@ def _build_source_tag_field(source_tag):
     # 3. Trim left/right whitespace
     # 4. Condense multiple spaces to one, and replace with underscore
     tag = source_tag.lower()
-    tag = re.sub(r'[^a-z0-9_ \:]', '', tag)
+    tag = re.sub(r'[^a-z0-9_ :]', '', tag)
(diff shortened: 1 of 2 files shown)


austin-denoble (approved):
Nice, this should be just as straightforward in the other clients since we handled it similarly. 🚢

pinecone/utils/user_agent.py
```diff
@@ -11,7 +11,7 @@ def _build_source_tag_field(source_tag):
     # 3. Trim left/right whitespace
     # 4. Condense multiple spaces to one, and replace with underscore
     tag = source_tag.lower()
-    tag = re.sub(r'[^a-z0-9_ ]', '', tag)
+    tag = re.sub(r'[^a-z0-9_ \:]', '', tag)
```
ssmith-pc, line 14: I think it's not necessary to escape the colon if it's used by itself like this

```python
>>> re.sub(r'[^a-z0-9_ :]', '', tag)
'my:tag'
```

ssmith-pc (approved):
🚀

PR #367, opened 2024-07-18
Implement deletion protection

jhamon:
## Problem

Implement deletion protection for python sdk

## Solution

- For the basic functionality, needed to update the create index and configure index actions
- The UX around the generated enum class, `DeletionProtection`, seemed extremely poor so I added a custom wrapper class `IndexModel` and incorporated it into the `describe_index` and `list_indexes` actions
- Update and add new tests
- Remove export of the generated IndexModel from the top-level `__init__.py` file.

This ended up being a much bigger lift than expected, but as a nice side effect we're no longer exposing generated objects in responses from these various actions. That creates some additional flexibility for future refactoring work.

## Type of Change

## Test Plan


pinecone/control/pinecone.py
```diff
@@ -291,7 +296,8 @@ def create_index(
             name="my_index",
             dimension=1536,
             metric="cosine",
-            spec=ServerlessSpec(cloud="aws", region="us-west-2")
+            spec=ServerlessSpec(cloud="aws", region="us-west-2"),
+            deletion_protection="enabled"
```
jhamon, line 300: This is docstrings code.

pinecone/control/pinecone.py
```diff
@@ -320,6 +327,11 @@ def create_index(
         def _parse_non_empty_args(args: List[Tuple[str, Any]]) -> Dict[str, Any]:
             return {arg_name: val for arg_name, val in args if val is not None}
 
+        if deletion_protection in ["enabled", "disabled"]:
+            dp = DeletionProtection(deletion_protection)
+        else:
+            raise ValueError("deletion_protection must be either 'enabled' or 'disabled'")
```
jhamon, line 333: The generated `DeletionProtection` class implements a validation, but the error message it emits is very confusing. So I implemented my own check on this here and in the configure method.

pinecone/control/pinecone.py
```diff
@@ -561,15 +575,28 @@ def configure_index(
 
         """
         api_instance = self.index_api
-        config_args: Dict[str, Any] = {}
+
+        if deletion_protection is None:
+            description = self.describe_index(name=name)
+            dp = DeletionProtection(description.deletion_protection)
```
jhamon, line 581: If the user does not specify a value for deletion_protection, we call describe_index to look up the current value. This is needed to avoid disabling deletion protection when somebody is trying to call this with `replicas` or `pod_type` only.

pinecone/control/pinecone.py
```diff
@@ -561,15 +575,28 @@ def configure_index(
...
         if replicas:
-            config_args.update(replicas=replicas)
-        configure_index_request = ConfigureIndexRequest(
-            spec=ConfigureIndexRequestSpec(pod=ConfigureIndexRequestSpecPod(**config_args))
-        )
-        api_instance.configure_index(name, configure_index_request=configure_index_request)
+            pod_config_args.update(replicas=replicas)
+
+        if pod_config_args != {}:
+            spec = ConfigureIndexRequestSpec(pod=ConfigureIndexRequestSpecPod(**pod_config_args))
```
jhamon, line 594: I hate the arrangement of this code. Blame the black formatter. I need to update the configs to shorten the line length to something more reasonable.

austin-denoble (approved):
LGTM, thanks again for flagging the issues around the `deletionProtection` enum early. Made it much easier to test things across the other clients. 🎉 🚢 

It might warrant a broader discussion at some point as to whether we want to completely replace generated types for things like responses in order to give us more flexibility and control over the interface, but that's probably a pretty big lift.

PR #369, opened 2024-07-19
Remove `openapi_config` property

jhamon:
## Problem

We want to remove the `openapi_config` config property because we don't want generated openapi stuff in the public interface. This is a holdover from the v2 days of `pinecone.init()` but has never been documented; I would be surprised if anyone is actually using it.

## Solution

- Convert deprecation warning into an exception
- Remove unit tests relying on this deprecated option

## Type of Change


pinecone/control/pinecone.py
```diff
@@ -214,9 +214,8 @@ def __init__(
             )
 
         if kwargs.get("openapi_config", None):
-            warnings.warn(
-                "Passing openapi_config is deprecated and will be removed in a future release. Please pass settings such as proxy_url, proxy_headers, ssl_ca_certs, and ssl_verify directly to the Pinecone constructor as keyword arguments. See the README at https://github.com/pinecone-io/pinecone-python-client for examples.",
-                DeprecationWarning,
+            raise Exception(
```
austin-denoble, line 217: Ah, so the part that checks for the kwarg of this is part of the underlying generated code implementation? That's annoying.

austin-denoble (approved):

pinecone/control/pinecone.py
jhamon: The whole bucket of kwargs gets passed around in various places, so to be sure nobody has formed a dependency on these internals (so we can safely refactor later) we need to explicitly check and throw right at the entry point.

PR #370, opened 2024-07-19
Add inference plugin and tests

jhamon:
## Problem

We want to ship inference as part of the core SDK

## Solution

Add a dependency on v1.0.2 of the inference plugin, which includes a recent fix to plugin installation for use with the GRPC client.

## Usage

The plugin should automatically be installed and available for usage.

```python
from pinecone import Pinecone

pc = Pinecone(api_key='key')

embedding_model = "multilingual-e5-large"
embeddings = pc.inference.embed(
    model=embedding_model,
    inputs=["The quick brown fox jumps over the lazy dog.", "lorem ipsum"],
    parameters={"input_type": "query", "truncate": "END"},
)
```

## Type of Change

## Test Plan

See tests passing.

### DIFF
--- a/.github/workflows/testing-integration.yaml
+++ b/.github/workflows/testing-integration.yaml
@@ -3,53 +3,27 @@ name: "Integration Tests"
   workflow_call: {}
 
 jobs:
-  # setup-index:
-  #   name: Setup proxyconfig test index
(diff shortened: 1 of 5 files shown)


austin-denoble (approved):
🚀

PR #372, opened 2024-07-24
[Fix] Fetch when vector id string contains spaces

jhamon:
## Problem

Some data operations fail when the vector id string contains a space.

```python
from pinecone import Pinecone

pc = Pinecone()
pc.fetch(ids=["id with string"]) # no results returned, even when vector exists
```

## Solution

The problem occurred due to the way spaces were being encoded as `+` instead of `%20` in url query params. The fix was a small adjustment to our code generation templates.

I added test coverage for upsert / query / fetch with various weird ids to make sure the change in encoding hasn't broken any other use cases that could pop up.

## Type of Change

### DIFF
--- a/tests/integration/data/seed.py
+++ b/tests/integration/data/seed.py
@@ -1,6 +1,7 @@
 from ..helpers import poll_fetch_for_ids_in_namespace
 from pinecone import Vector
 from .utils import embedding_values
+import itertools
 
 
 def setup_data(idx, target_namespace, wait):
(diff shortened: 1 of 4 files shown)


tests/integration/data/seed.py
```diff
@@ -43,3 +44,82 @@ def setup_list_data(idx, target_namespace, wait):
...
+    unwanted_interpolation = [
+        "$HOME",
+        "$ENV{'HOME'}",
+        "%d",
+        "%s",
+        "%n",
+        "%x",
+        "{0}",
+    ]
+    ids.extend(unwanted_interpolation)
```
aulorbe, line 115: total nit, but it just be simpler to do something like `return booleanish + script_injection + unwanted_interpolation`, instead of calling `ids.extend` mult times?

aulorbe (approved):

PR #373, opened 2024-07-24
[CI] Publish doc updates after each release

jhamon:
## Problem

We want documentation to automatically stay up to date after every release

## Solution

Trigger the docs publish workflow after a new artifact is released to pypi

## Type of Change


austin-denoble (approved):
Thanks!

Issue #378, opened 2024-08-01
circular dependency issue involving the pinecone-client and pinecone-plugin-inference packages


msimc:
### Is this a new bug in the Pinecone Python client?

- [X] I believe this is a new bug in the Pinecone Python Client
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Hello Pinecone Team,

I've encountered a circular dependency issue involving the pinecone-client and pinecone-plugin-inference packages. Specifically, the pinecone-client package lists pinecone-plugin-inference as a required dependency, and pinecone-plugin-inference in turn depends on pinecone-client. This creates a circular dependency loop, which can cause issues during package installation and dependency resolution.

Details:

pinecone-client Version: 5.0.0
pinecone-plugin-inference Version: 1.0.2
Installation Method: Using pip in a virtual environment
Issue Description: During the dependency resolution process, a circular dependency is detected between pinecone-client and pinecone-plugin-inference. This is causing installation conflicts and difficulties in managing these packages in a clean virtual environment.
Steps to Reproduce:

Create a new virtual environment.
Attempt to install pinecone-client and observe the dependency resolution process.
Expected Behavior:

The packages should install without conflicts, and all dependencies should resolve correctly.
Actual Behavior:

A circular dependency is detected, potentially causing installation issues and preventing a clean setup.
Request:
Could you please advise on how to handle this situation or if there are plans to address this dependency loop in future releases? Any guidance or workaround to resolve this would be greatly appreciated.

Thank you for your attention to this matter.

Best regards,
[Your Name]

### Expected Behavior

A circular dependency is detected, potentially causing installation issues and preventing a clean setup.

### Steps To Reproduce

Create a new virtual environment.
Attempt to install pinecone-client and observe the dependency resolution process.

### Relevant log output

```shell
[Running] python -u "c:\Users\Martin\Repositories\PDF Version Checker\All_Info\generate_full_dependency_tree.py"
Circular dependency detected: pinecone-client
Dependency information saved to c:\Users\Martin\Repositories\PDF Version Checker\All_Info\generate_full_dependency_tree_2024-08-01.txt

[Done] exited with code=0 in 40.363 seconds
```

### Environment

```markdown
- OS:WIndows 11
- Python:3.12.4
- 
PINECONE_INDEX_NAME=pineconetestindex
PINECONE_DIMENSION=384
PINECONE_REGION=us-east-1
```

### Additional Context

import os
import sys
[...]

PR #379, opened 2024-08-01
Adjusting inference plugin dependency

jhamon:
## Problem

Issue #378 raised a concern about a circular dependency relationship between the `pinecone-client` and `pinecone-plugin-inference` plugin which implements new Inference API functionality.

## Solution

We have updated the plugin to remove it's declared dependency on pinecone-client, and now need to update the pinecone-client to consume that update.

## Type of Change


pyproject.toml
```diff
@@ -68,7 +68,7 @@ lz4 = { version = ">=3.1.3", optional = true }
 protobuf = { version = "^4.25", optional = true }
 protoc-gen-openapiv2 = {version = "^0.0.1", optional = true }
 pinecone-plugin-interface = "^0.0.7"
-pinecone-plugin-inference = ">=1.0.2"
+pinecone-plugin-inference = ">=1.0.3"
```
ssmith-pc, line 71: Can we change it to the caret style ^1.0.3? That will restrict it to ">=1.0.3, <2.0.0". If inference is doing a major version upgrade, then it should probably result in a major version upgrade for the main client

ssmith-pc (approved):

ssmith-pc (approved):

PR #383, opened 2024-08-12
Rename package from pinecone-client to pinecone

jhamon:
## Problem

We want to rename this package from `pinecone-client` to `pinecone`.

## Solution

- Update `pyproject.toml` to refer to the new name `pinecone`
- Find and replace references to old package name in README and docstrings
- Remove nightly publish job that would need to be updated but has been broken for a long time for other reasons

## Type of Change

- [x] None of the above: We plan to publish a wrapper package under the name `pinecone-client`. So this shouldn't be a breaking change for end users consuming updates via the old name.

## Test Plan


rohanshah18:
What's the reason for renaming the package?

rohanshah18 (approved):
What's the reason for renaming the package?

austin-denoble (approved):
LGTM, thanks again for getting this over the line.

jhamon:
> What's the reason for renaming the package?

We went through a [long process](https://github.com/pypi/support/issues/2432) this year to get control of the package name `pinecone` which was previously being squatted by an inactive project. Pinecone is our trademark, so makes sense that we should have the name since it wasn't really being used. Now we have to do some work to migrate over everything (docs, etc) but don't want to break compatibility with anything else out there referencing the old name.

PR #385, opened 2024-08-16
Add has_index()

rohanshah18:
## Problem

Currently, users can not easily check if an index exists or not. The .names() iterator on the list_indexes response object is not super discoverable. 

## Solution
Add has_index() which internally calls describe_index() and returns true if the index exists, false otherwise.

## Type of Change

## Test Plan

Added integration tests.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207972900637843


pinecone/control/pinecone.py
```diff
@@ -737,12 +764,12 @@ def Index(self, name: str = "", host: str = "", **kwargs):
 
         pc = Pinecone(api_key=api_key)
         pc.create_index(
-            name='my-index',
+            name='my_index',
```
rohanshah18, line 767: - is not allowed in index naming

pinecone/control/pinecone.py
```diff
@@ -737,12 +764,12 @@ def Index(self, name: str = "", host: str = "", **kwargs):
...
             dimension=1536,
             metric='cosine',
             spec=ServerlessSpec(cloud='aws', region='us-west-2')
         )
-        index = pc.Index('my-index')
+        index = pc.Index('my_index')
```
rohanshah18, line 772: - is not allowed in index naming

pinecone/control/pinecone.py
```diff
@@ -544,6 +544,33 @@ def describe_index(self, name: str):
...
+        ```python
+        from pinecone import Pinecone
+
+        client = YourClass()  # Initialize your class instance
+        exists = client.has_index("my_index_name")
+
+        if exists:
+            print("The index exists.")
+        else:
+            print("The index does not exist.")
```
jhamon, line 564: I would make this example code like this to be more similar to the style in other examples

```python
import os
from pinecone import Pinecone

api_key = os.environ.get("PINECONE_API_KEY")
pc = Pinecone(api_key=api_key)

if pc.has_index("my_index_name"):
    print("The index exists")
else:
    print("The index does not exist")
```

I think the way this is most likely to be used is in notebooks and examples saying something like this, with no intermediate variables created:

```python
if !pc.has_index("example_index"):
    pc.create_index(... args here)
```

pinecone/control/pinecone.py
```diff
@@ -544,6 +544,33 @@ def describe_index(self, name: str):
...
+        ```
+        """
+
+        try:
+            self.index_api.describe_index(name)
+            return True
+        except Exception as e:
+            return False
```
jhamon, line 572: This should work most of the time, but now that I'm thinking about it more I realized that exceptions can occur that should not be ignored. For example, a 500 error coming from the backend should throw rather than be ignored. Or if you try calling with an incorrect api key, that's a 401 unauthorized that the user should see an error about.

It would probably be simpler to call `list_indexes` and check if the index is in the list instead of ignoring exceptions.

jhamon (approved):
I left some feedback to fix, but overall great job putting together your first PR in python :)

pinecone/control/pinecone.py
rohanshah18: Agreed, this is definitely the downside of the former approach:
> For example, a 500 error coming from the backend should throw rather than be ignored. Or if you try calling with an incorrect api key, that's a 401 unauthorized that the user should see an error about.

jhamon (approved):
Looks great! 🚢

PR #386, opened 2024-08-22
Add bulk import

jhamon:
## Problem

Implement the following new methods:

- `start_import`
- `describe_import`
- `list_imports`
- `cancel_import`

#### Code generation changes

Since these features are in prerelease, they only exist in the spec for the upcoming 2024-10 API version. This required me to make modifications to the codegen script that is now run as:

```
./codegen/build-oas.sh 2024-07 false && ./codegen/build-oas.sh 2024-10 true
```

The second boolean argument is used to tell the codegen script whether the generated code should be stored in a new `pinecone/core_ea` subpackage. In the future we should probably do more to hide this complexity from the developer, but for now it is good enough.

#### Code organization

For the bespoke bits of the implementation that wrap the generated code, I have put them into a new class, `ImportFeatureMixin`, that the `Index` class inherits from. These functions could have all been implemented directly in the `Index` class, but I thought it a bit tidier to segregate these into a separate spot than just dump everything into one giant file.

#### Overridden repr representation on generated objects

The default print output in the generated classes comes from pprint and it looks quite poor for large objects. So I installed overrides that dump the objects into a formatted json style instead. I had previously done something similar for describe_index, etc, methods, so for this PR it was just a matter of cleaning up that logic a bit and moving it somewhere it could be reused.

So far, I haven't tweaked the generated classes to do this approach across the board because it doesn't work well for long arrays of vector values.

## Type of Change

## Test Plan

Manual testing with a dev release is in this [demo notebook](https://colab.research.google.com/drive/1W3OhMDG1yW2rdwx-ZulYH847m9R_IUuK#scrollTo=gGvVbfkYNz61
)


.github/workflows/alpha-release.yaml
```diff
@@ -24,22 +24,22 @@ on:
...
   pypi:
     uses: './.github/workflows/publish-to-pypi.yaml'
-    needs: 
-      - unit-tests
-      - integration-tests
-      - dependency-tests
+    # needs: 
+    #   - unit-tests
+    #   - integration-tests
+    #   - dependency-tests
```
jhamon, line 42: I disabled these to speed up the process of creating dev releases. But will uncomment before merging. Eventually some sort of menu toggle for skipping tests would be nice, but I didn't want to yak shave on that too much.

pinecone/core/openapi/shared/model_utils.py
```diff
@@ -738,12 +740,12 @@ def get_simple_class(input_value):
...
-    #    # isinstance(datetime_instance, date) == True
-    #    return datetime
-    # elif isinstance(input_value, date):
-    #    return date
+    elif isinstance(input_value, datetime):
+        # this must be higher than the date check because
+        # isinstance(datetime_instance, date) == True
+        return datetime
+    elif isinstance(input_value, date):
+        return date
```
jhamon, line 748: The date/datetime changes in these generated files reflect some adjustments I had to make in the template code to get correct handling of dates in API responses since the bulk import feature introduces date fields for the first time. Previously we had all of this commented out as a quick-and-dirty fix for an unrelated issue where untyped user metadata strings stored with vectors were sometimes coerced into datetime when fetched if the content looked date-ish.

There's a unit test covering that case to make sure we haven't regressed on that issue while still enabling us to interact with datetime objects when that is our intention.

tests/unit/data/test_bulk_import.py
```diff
@@ -0,0 +1,172 @@
+import pytest
```
jhamon, line 1: Here's where new unit tests were added for the main functionality

pinecone/config/openapi.py
```diff
@@ -24,6 +24,8 @@ def build(cls, api_key: str, host: Optional[str] = None, **kwargs):
         openapi_config.host = host
         openapi_config.ssl_ca_cert = certifi.where()
         openapi_config.socket_options = cls._get_socket_options()
+        openapi_config.discard_unknown_keys = True
```
jhamon, line 27: Discovered this along the way. Seems like a better default behavior than erroring when unexpected data is returned.

pinecone/data/features/bulk_import.py
```diff
@@ -0,0 +1,193 @@
+from enum import Enum
```
jhamon, line 1: Majority of the new code is in this file.

aulorbe:
Total nit that obviously you can ignore if you want, but maybe we should change the title of this PR from "Early access bulk import" to "Add early access bulk import", just so users later on know that this PR added the functionality (instead of iterated/removed/etc.)

CONTRIBUTING.md
```diff
@@ -142,3 +142,26 @@ Hello, from your virtualenv!
...
 If you experience any issues please [file a new issue](https://github.com/pinecone-io/pinecone-python-client/issues/new).
+
+
+## Consuming API version upgrades
+
+These instructions can only be followed by Pinecone employees with access to our private APIs repository.
+
+Prerequisites:
+- You must be an employee with access to private Pinecone repositories
+- You must have [Docker Desktop](https://www.docker.com/products/docker-desktop/) installed and running. Our code generation script uses a dockerized version of the openapi CLI.
```
aulorbe, line 153: Nit: you might want to capitalize OpenAPI

Makefile
```diff
@@ -11,7 +11,7 @@ develop:
 
 test-unit:
 	@echo "Running tests..."
-	poetry run pytest --cov=pinecone --timeout=120 tests/unit
+	poetry run pytest --cov=pinecone --timeout=120 tests/unit -s -vv
```
aulorbe, line 14: Just making sure: you want to permanently add these flags here? I just know that they're usually added for debugging, so want to make sure they weren't left accidentally

codegen/build-oas.sh
```diff
@@ -21,9 +33,9 @@ update_apis_repo() {
 update_templates_repo() {
 	echo "Updating templates repo"
 	pushd codegen/python-oas-templates
-		git fetch
-		git checkout main
-		git pull
+		# git fetch
```
aulorbe, line 36: OOC why'd you remove these?

pinecone/config/openapi.py
aulorbe: Looks sweet

pinecone/data/features/bulk_import.py
```diff
@@ -0,0 +1,193 @@
...
+            )
+
+    def start_import(
+        self,
+        uri: str,
+        integration_id: Optional[str] = None,
+        error_mode: Optional[Literal["CONTINUE", "ABORT"]] = "CONTINUE",
+    ) -> StartImportResponse:
+        """Import data from a storage provider into an index. The uri must start with the scheme of a supported
+        storage provider. For buckets that are not publicly readable, you will also need to separately configure
```
aulorbe, line 53: Maybe so users have a better understanding of what we mean when we say "storage provider," you write "e.g. "S3"" or something?

pinecone/data/features/bulk_import.py
```diff
@@ -0,0 +1,193 @@
...
+        a storage integration and pass the integration name.
```
aulorbe, line 54: Should this read "integration ID"? (not name)

pinecone/data/features/bulk_import.py
```diff
@@ -0,0 +1,193 @@
...
+
+        Examples:
+            >>> from pinecone import Pinecone
+            >>> index = Pinecone().Index('my-index')
+            >>> index.start_import(uri="s3://bucket-name/path/to/data.parquet")
+            { id: "1" }
+
+        Args:
+            uri (str): The URI of the data to import. The URI must start with the scheme of a supported storage provider.
+            integration (Optional[str], optional): If your bucket requires authentication to access, you need to pass the name of your storage integration using this property. Defaults to None.
```
aulorbe, line 64: Should this be integration ID not name?

pinecone/data/features/bulk_import.py
```diff
@@ -0,0 +1,193 @@
...
+            error_mode: Defaults to "CONTINUE". If set to "CONTINUE", the import operation will continue even if some
```
aulorbe, line 65: idk if you want to do this, but in the TS client I lowercase whatever they input just to avoid imports failing b/c users put "Continue" vs "CONTINUE"

pinecone/data/features/bulk_import.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        Returns a generator that yields each import operation. It automatically handles pagination tokens on your behalf so you can
+        easily iterate over all results. The `list_imports` method accepts all of the same arguments as list_imports_paginated
+
+        ```python
+        for op in index.list_imports():
+            print(op)
+        ```
+
+        You can convert the generator into a list by wrapping the generator in a call to the built-in `list` function:
```
aulorbe, line 99: noice

pinecone/data/features/bulk_import.py
```diff
@@ -0,0 +1,193 @@
...
+
+        ```python
+        operations = list(index.list_imports())
+        ```
+
+        You should be cautious with this approach because it will fetch all operations at once, which could be a large number
+        network calls and a lot of memory to hold the results.
```
aulorbe, line 106: I think you're missing an "of" here (large number OF network calls)

aulorbe (approved):
lgtm! just some nits/language tweaks

codegen/build-oas.sh
jhamon: Nice catch, I didn't mean to leave these commented. I disabled the pull because I was experimenting with some local changes to get the datetime parsing properly.

pinecone/data/features/bulk_import.py
jhamon: Currently only S3 is supported, but I don't want to write that explicitly because I think it will quickly go out of date. The error message the API sends back is relatively helpful if you get this wrong.

CONTRIBUTING.md
```diff
@@ -142,3 +142,26 @@ Hello, from your virtualenv!
 ```
 
 If you experience any issues please [file a new issue](https://github.com/pinecone-io/pinecone-python-client/issues/new).
+
+
+## Consuming API version upgrades
```
austin-denoble, line 147: **thought:** We should definitely put something like this in other repos where we're using a `/codegen/` submodule in other repos, we had someone external asking about codegen in the Rust codebase. Would be good to be explicit until we've got the specs published, plus for better documenting how the codegen actually works in each repo.

codegen/build-oas.sh
```diff
@@ -3,9 +3,21 @@
 set -eux -o pipefail
 
 version=$1 # e.g. 2024-07
-modules=("control" "data")
+is_early_access=$2 # e.g. true
+
+# if is_early_access is true, add the "ea" module
+if [ "$is_early_access" = "true" ]; then
```
austin-denoble, line 9: **thought:** Clever way of going about this, nice. 👍

codegen/build-oas.sh
```diff
@@ -73,13 +83,20 @@ generate_client() {
 	docker run --rm -v $(pwd):/workspace openapitools/openapi-generator-cli:v5.2.0 generate \
 		--input-spec "/workspace/$oas_file" \
 		--generator-name python \
-		--config "/workspace/$openapi_generator_config" \
+		--additional-properties=packageName=$package_name,pythonAttrNoneIfUnset=true \
 		--output "/workspace/${build_dir}" \
 		--template-dir "/workspace/$template_dir"
 
+	# Hack to prevent coercion of strings into datetimes within "object" types while still
+	# allowing datetime parsing for fields that are explicitly typed as datetime
```
austin-denoble, line 91: Cool, does this replace the templating we used to have to do to fix the previous `datetime` problem?

pinecone/config/openapi.py
austin-denoble: Good catch

pinecone/core/openapi/shared/model_utils.py
austin-denoble: Answers my previous question about the templates, thanks.

pinecone/data/index.py
```diff
@@ -64,7 +66,7 @@ def parse_query_response(response: QueryResponse):
     return response
 
 
-class Index:
+class Index(ImportFeatureMixin):
```
austin-denoble, line 69: Very cool, I really like the mixin approach.

pinecone/data/features/bulk_import.py
austin-denoble: Since we bumped the version of Node and some of our TypeScript config stuff, we may be able to use generators in the TypeScript codebase.  🤔 

I had written one for the `List` endpoint but ended up not shipping it.

austin-denoble (approved):
LGTM, really nice work getting all of this in place. Very thoughtful approach. Do you think it would make sense doing something similar in other repos? I recently created a release candidate branch to work against in Go.

Really nice work, Jen.

codegen/build-oas.sh
jhamon: Yeah, exactly. I rolled back the old changes to the templates and replaced with this. It's only this one spot where the datetime stuff was problematic.

PR #387, opened 2024-08-28
Improve output from list/describe actions on indexes and collections

jhamon:
## Problem

When running in notebooks, output from the `list_indexes` and `list_collections` commands created confusion because the `__repr__` representation for these objects showed an object with a top-level key that implied the response should be interacted with like a dictionary. This expectation contradicts with the way the `__iter__` implementations on these results objects are set up to enable iterating over results without drilling down. 

The origin of the complexity here is that the backing APIs used to return simple arrays of names that could be easily iterated over, and earlier this year they become more fleshed out responses. Returning more data is useful in some situations, but in trying to smooth out the impact of the API change and maintain a similar way of interacting with the results, we accidentally opened up this inconsistency in the experience.

## Solution

- For these actions, migrate from output by `pprint.pformat` to `json.dumps`. For deeply nested objects, this produces a result that is easier to read.
- Stop sorting keys alphabetically. The way they are returned from the API makes the most sense (name first).
- Remove top-level keys from the printed output, which created wrong expectations about how to interact with the results object. Now list_indexes looks like an array, and you index into it like an array. Ditto for collections.

### Before

Screenshot 2024-08-28 at 1 38 27 PM

Collections output

### After

Screenshot 2024-08-28 at 1 29 00 PM

Screenshot 2024-08-28 at 1 42 05 PM

Screenshot 2024-08-28 at 3 28 44 PM

## Type of Change

- [X] None of the above: UX improvement, but should be no functional change

### DIFF
--- a/pinecone/control/__init__.py
+++ b/pinecone/control/__init__.py
@@ -1,4 +1,5 @@
 from .pinecone import Pinecone
 
 from .repr_overrides import install_repr_overrides
+
 install_repr_overrides()
--- a/pinecone/control/__init__.py
+++ b/pinecone/control/__init__.py
@@ -1,1 +1,4 @@
 from .pinecone import Pinecone
+
+from .repr_overrides import install_repr_overrides
+install_repr_overrides()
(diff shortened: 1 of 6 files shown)


rohanshah18 (approved):
Looks nice, LGTM!

PR #393, opened 2024-09-26
Upgrade protobuf from v4 to v5: 5x Upsert throughput

daverigby:
Upgrade the protobuf dependancy from v4 (4.25) to v5 (5.28). This
appears to have significantly faster protobuf encoding - I see a 4.5x - 5x
inprovement in Upsert throughput on a given EC2 machine (i3.xlarge)
for large batches (~300) of high dimensionality vectors (1536):

Before:

    Performing Populate phase               1675770/138364198 ╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   1% 0:35:34 44:17:17
      Records/sec: 785.2

After:

    Performing Populate phase               1531830/138364198 ╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   1% 0:07:07 10:35:48
      Records/sec: 3584.4

I haven't dug into the exact details, but the profile is quite
different - the python frames performing type checking are no longer
present, so I assume they have been optimised, perhaps pushed to
native code?

Before profile:

protobuf_v4

After profile:

protobuf_v5

## Type of Change

- [x] None of the above: Dependency upgrade.

## Test Plan


jhamon:
Looks good, thanks! Changing dependencies is a breaking change, but it should go out soon when we release new SDKs for the Oct 15 API version.

PR #397, opened 2024-10-16
Update to pinecone-plugin-inference=2.0.0

ssmith-pc:
## Problem
Updates `pinecone-plugin-inference` to new major version `2.0.0`.

From the 2.0.0 release notes:

### Reusing top-level exceptions from `pinecone` client
Exceptions in the Pinecone Inference SDK have been reworked to throw the top-level exceptions declared in the Pinecone Python SDK as opposed to plugin-specific duplications of those exceptions, which was confusing to users. 

#### Old
Previously, exceptions in the Python Inference SDK were redeclared/duplicated in an obscure package, leading to a poor user experience when working with what should otherwise be standard Pinecone exceptions. 

For example, a user who wanted to catch and handle `PineconeApiException` would have to know to enter:
```python
from pinecone_plugins.inference.core.client.exceptions import PineconeApiException
```

#### New
Now, the Inference SDK reuses exceptions from the top-level Pinecone SDK, allowing the user to simply enter:
```python
from pinecone.exceptions import PineconeApiException 
```

## Solution

## Type of Change

## Test Plan


austin-denoble (approved):
🚢

PR #398, opened 2024-10-18
[Feat] asyncio for grpc data interactions

jhamon:
## Problem

## Solution

## Type of Change

## Test Plan


pinecone/grpc/query_results_aggregator.py
```diff
@@ -0,0 +1,178 @@
...
+            # if you read results twice due to the heap being emptied when constructing
+            # the ordered results.
+            raise ValueError("Results have already been read. Cannot add more results.")
+
+        matches = results.get("matches", [])
+        ns: str = results.get("namespace", "")
+        self.usage_read_units += results.get("usage", {}).get("readUnits", 0)
+
+        if len(matches) == 0:
+            raise QueryResultsAggregationEmptyResultsError(ns)
```
Noam-Barbiru, line 151: We shouldn't panic if a namespace doesn't return results. It is ok to return zero results, like in a case where the filter filtered out all the vectors in the namespace.

jhamon: Thanks for calling this out. Jeff originally asked that we fail when namespace is incorrect, but that seems indistinguishable from the case where results are empty due to filtering. So I'm going to take this out.

asvishnyakov:
@jhamon Do you have plans to release this? I'm highly interested in with asyncio version of index

PR #401, opened 2024-10-21
Refactoring and updates for 2024-10 API spec

jhamon:
## Problem

Needed to consume the 2024-10 API spec.

## Solution

The spec updates involved many changes to operation ids. This should not touch the user experience, but does mean this diff is somewhat larger than usual due to the large number of tests for existing functionality that needed to be adjusted for the new names:
- upsert > upsert_vector
- fetch > fetch_vectors
- query > query_vectors
- delete > delete_vectors
- list > list_vectors
- start_import > start_bulk_import
- describe_import > describe_bulk_import
- etc

#### Cleanup prerelease code no longer needed
- Deleted `pinecone/core_ea/` which was the home of some code generated off of the 2024-10 spec when bulk import functionality was in a pre-release state. We no longer need this. That's why this PR has 21k lines removed.

#### Changes to generation process
- Placed a copy of classes such as `ApiClient`, `Endpoint`, `ModelNormal`, etc which were previously generated into a new folder, `pinecone/openapi_support`. Even though these used to be generated, they contain no dynamic content. So taking them out of the generation process should make it significantly easier to work on core improvements to the UX and performance of the generated code, since the source of truth will no longer be a mustache file (FML)
- Updated the codegen script `./codegen/build-oas.sh 2024-10 false` to delete instead of de-dupe copies of shared classes like `ApiClient`, and edit model+api files that continue to be generated to find the classes they need in the new `openapi_support` folder.

#### mypy fixes
- I needed to make adjustments in some `openapi_support` classes to satisfy the mypy type checker (`pinecone/core` has been ignored in the past, and our goal should be to eventually have 100% of code typechecked).
- Wrote some initial unit tests to characterize some of the behavior in `ApiClient`. The tests were to help give me more confidence I wasn't breaking something along the way while making adjustments for mypy.
- Added a dev dependency on a package with types for datetime

## Type of Change


pinecone/core/openapi/db_data/api/bulk_operations_api.py
```diff
@@ -89,63 +89,43 @@ def __cancel_import(self, id, **kwargs):
...
             return self.call_with_http_info(**kwargs)
 
-        self.cancel_import = _Endpoint(
+        self.cancel_bulk_import = _Endpoint(
             settings={
                 "response_type": ({str: (bool, dict, float, int, list, str, none_type)},),
                 "auth": ["ApiKeyAuth"],
                 "endpoint_path": "/bulk/imports/{id}",
-                "operation_id": "cancel_import",
+                "operation_id": "cancel_bulk_import",
```
aulorbe, line 97: I thought we were _not_ using the word "bulk" in the clients anymore wrt to this endpoint? https://pinecone-io.slack.com/archives/C05MYEX7T2L/p1728935415605249?thread_ts=1728935400.761069&cid=C05MYEX7T2L

cc @austin-denoble

aulorbe: (Looks like it's not used in `apis` either: e.g. https://github.com/pinecone-io/apis/blob/main/static/2024-10/db_data_2024-10.oas.yaml#L76)

aulorbe (changes requested):
I think we need to remove "bulk" from the import language

haruska (approved):

pinecone/core/openapi/db_data/api/bulk_operations_api.py
austin-denoble: This is generated code, so not really publicly exposed. I think this is fine, there's a similar story in other SDKs since the spec contains "bulk" in a few places.

codegen/build-oas.sh
```diff
@@ -13,7 +13,7 @@ if [ "$is_early_access" = "true" ]; then
 	template_dir="codegen/python-oas-templates/templates5.2.0"
 else
 	destination="pinecone/core/openapi"
-	modules=("control" "data")
+	modules=("db_control" "db_data")
```
rohanshah18, line 16: Is inference going to be a part of the plugin?

rohanshah18 (approved):
Left a question about inference. Otherwise it looks good, LGTM!

codegen/build-oas.sh
jhamon: For now, inference stuff comes in via plugin which is why you don't see anything about it here.

PR #402, opened 2024-10-22
Detect plugins for Index and IndexGRPC classes

jhamon:
## Problem

Plugins not detected for Index/IndexGRPC classes

## Solution

- Copy and modify installation method in pinecone.py used for `Pinecone` and `PineconeGRPC` classes.
- For IndexGRPC, pass in a stub function instead of 

## Type of Change

## Test Plan

To really validate this will probably require making a dev build release and installing alongside some actual plugins.

### DIFF
--- a/pinecone/data/index.py
+++ b/pinecone/data/index.py
@@ -1,11 +1,11 @@
 from tqdm.autonotebook import tqdm
 
+import logging
 from typing import Union, List, Optional, Dict, Any
 
 from pinecone.config import ConfigBuilder
 
 from pinecone.core.openapi.shared import API_VERSION
-from pinecone.core.openapi.data.models import SparseValues
 from pinecone.core.openapi.data import ApiClient
 from pinecone.core.openapi.data.models import (
     FetchResponse,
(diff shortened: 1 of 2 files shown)


pinecone/data/index.py
```diff
@@ -22,12 +22,22 @@
...
-from .features.bulk_import import ImportFeatureMixin
 from pinecone.core.openapi.data.api.data_plane_api import DataPlaneApi
-from ..utils import setup_openapi_client, parse_non_empty_args
+from ..utils import (
+    setup_openapi_client,
+    parse_non_empty_args,
+    build_plugin_setup_client,
+    validate_and_convert_errors,
+)
+from .features.bulk_import import ImportFeatureMixin
```
jhamon, line 34: Most of these import changes are just cosmetic

austin-denoble (approved):
LGTM and makes sense. I'll take the dev build and get it tested with a plugin before we merge and move on, but thanks for getting this together.

pinecone/data/index.py
```diff
@@ -103,6 +111,23 @@ def __init__(
...
+        self._load_plugins()
+
+    def _load_plugins(self):
+        """@private"""
+        try:
+            # I don't expect this to ever throw, but wrapping this in a
+            # try block just in case to make sure a bad plugin doesn't
+            # halt client initialization.
+            openapi_client_builder = build_plugin_setup_client(
+                config=self.config,
```
austin-denoble, line 123: I think this needs to be `self._config`. Currently seeing this error when it attempts to apply plugins to `Index`:

```
ERROR    pinecone.data.index:index.py:129 Error loading plugins in Index: 'Index' object has no attribute 'config'
```

PR #404, opened 2024-10-24
[Chore] Improve test flakes

jhamon:
## Problem

I frequently see recurring test flakes, most often at the cleanup stage of `test_configure_index_with_deletion_protection` because the index cannot be deleted while still in an "upgrading" state. The index is in that state while configuration changes are being applied.

## Solution

- In deletion protection test, wait for the index to be ready before attempting the delete.
- In dependency tests which do a basic sanity test, loosen the assertions. We're not really intending to test the backend functionality with these tests; we just want to make sure network calls are successfully made with a variety of different grpc / protobuff / etc dependency versions.
- Remove some assertions that rely on non-deterministic behavior on the backend. E.g. how long it takes for a serverless index to be considered ready.
- Check for deletion protection status before trying to delete.
- Don't fail the job when cleanup fails; even if we re-run the entire job, there will still be an orphaned index from a previous run that needs to be cleaned up. Best to just attempt to delete, then let the nightly cleanup job garbage collect anything left over.

## Type of Change

### DIFF
--- a/scripts/cleanup-all.py
+++ b/scripts/cleanup-all.py
@@ -2,6 +2,7 @@ import os
 import re
 from pinecone import Pinecone
 from datetime import datetime, timedelta
+import time
 
 
 def delete_everything(pc):
@@ -16,6 +17,30 @@ def delete_everything(pc):
     for index in pc.list_indexes().names():
         try:
             print("Deleting index: " + index)
+            desc = pc.describe_index(index)
+
+            # Check whether index can be deleted
+            if desc.deletion_protection == "enabled":
+                pc.configure_index(index, deletion_protection="disabled")
+
+            # Wait for index to be ready before deleting
+            ready_to_delete = False
+            max_wait = 60
+            time_waited = 0
+            while not ready_to_delete:
+                desc = pc.describe_index(index)
+                if desc.status.state == "Ready":
+                    ready_to_delete = True
+                    break
+                else:
+                    print("Index is not ready yet. Waiting for 2 seconds.")
+                    time.sleep(2)
+                    time_waited += 2
+
+                if time_waited > max_wait:
+                    print(f"Timed out waiting for index {index} to be ready")
+                    break
+
             pc.delete_index(index)
         except Exception as e:
             print("Failed to delete index: " + index + " " + str(e))
--- a/.github/actions/cleanup-all/action.yml
+++ b/.github/actions/cleanup-all/action.yml
@@ -5,12 +5,16 @@ inputs:
   PINECONE_API_KEY:
     description: 'The Pinecone API key'
     required: true
+  DELETE_ALL:
+    description: 'Delete all indexes and collections'
+    required: false
+    default: 'false'
 
 runs:
   using: 'composite'
   steps:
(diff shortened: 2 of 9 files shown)


rohanshah18 (approved):
LGTM

scripts/cleanup-all.py
```diff
@@ -16,6 +17,30 @@ def delete_everything(pc):
...
+                desc = pc.describe_index(index)
+                if desc.status.state == "Ready":
+                    ready_to_delete = True
+                    break
+                else:
+                    print("Index is not ready yet. Waiting for 2 seconds.")
+                    time.sleep(2)
+                    time_waited += 2
+
+                if time_waited > max_wait:
```
haruska, line 40: small nit: because of logic ordering, you end up sleeping for 2 seconds before failing on the last iteration.

scripts/cleanup-all.py
```diff
@@ -16,6 +17,30 @@ def delete_everything(pc):
...
+
+            # Wait for index to be ready before deleting
+            ready_to_delete = False
+            max_wait = 60
+            time_waited = 0
+            while not ready_to_delete:
+                desc = pc.describe_index(index)
+                if desc.status.state == "Ready":
+                    ready_to_delete = True
+                    break
```
haruska, line 34: why both a break and boolean flag?  could just loop/break?

scripts/cleanup-all.py
```diff
@@ -16,6 +17,30 @@ def delete_everything(pc):
...
+            if desc.deletion_protection == "enabled":
+                pc.configure_index(index, deletion_protection="disabled")
+
+            # Wait for index to be ready before deleting
+            ready_to_delete = False
+            max_wait = 60
+            time_waited = 0
+            while not ready_to_delete:
+                desc = pc.describe_index(index)
+                if desc.status.state == "Ready":
```
haruska, line 32: I believe you can delete in other states as well.

haruska (approved):

scripts/cleanup-all.py
jhamon: You would think, wouldn't you? Maybe this is true in some cases, but a lot of flakes seen in the past are about trying to delete while the index is "upgrading". I.e. if you called configure index too recently, your delete request will be rejected.

scripts/cleanup-all.py
jhamon: This is what comes from outsourcing too much typing to github copilot 🤣

scripts/cleanup-all.py
austin-denoble: I actually changed all my integration tests to check both `status.state` and `status.ready`. There are situations where they are not in sync, and the index is marked as Ready but then the state is something else.

I'd maybe recommend checking both just to be thorough, it's helped with some of my flakiness. I think @aulorbe reported this to the db team at some point.

austin-denoble (approved):
LGTM, this also seems in line with some of the approaches we've had to take in other clients to massage some of these issues, specifically around cleanup.

I basically just retry the cleanup delete call a set number of times before bailing in Go.

PR #405, opened 2024-10-25
WIP on threadpool impl of query_namespaces

jhamon:
## Problem

I need a non-async interface providing a similar functionality as the `query_namespaces` currently implemented in `GRPCIndexAsyncio`.

### Background info:
- The `multiprocessing` ThreadPool is currently setup during the instantiation of `Index`. The constructor takes a param for number of `pool_threads` with a default value of 1.
- When calling query with optional kwarg, `async_req=True`, the underlying (generated from openapi spec) code will execute the request using `pool.apply_async`. The result of these calls is an `ApplyResult[QueryResult]`. To get the results, from these we call `.get()` on the `ApplyResult`.

### New `query_namespaces` on `Index`:
- Copy some basic edge case stuff from the asyncio implementation. E.g. must have namespaces arg.
- For each namespace, call `query` with `async_req=True`
- Gather the resulting `ApplyResults` into a results array
- Pass these results into `QueryResultsAggregator`, recently implemented and tested in https://github.com/pinecone-io/pinecone-python-client/pull/398
- It's expected this will be slower than the Asyncio implementation because we can't begin building the combined results heap until all requests have completed
- Added some retry logic for requests made with thread pools

### Todo:
- Performance testing

### Usage

```python
from pinecone import Pinecone
import random

pc = Pinecone(
    api_key=os.environ['PINECONE_API_KEY'],
)

index = pc.Index(
    host="https://indexhost/",
    pool_threads=10
)

query_vec = [random.random()] * dimension

combined_results = index.query_namespaces(
    vector=query_vec,
    namespaces=["ns1", "ns2", "ns3", "ns4"],
    include_values=True,
    include_metadata=True,
    filter={"publication_date": {"$eq":"Last3Months"}},
    top_k=100
)
```

### Dev build for manual testing
- `pip install pinecone==6.0.0.dev3`

## Type of Change

## Test Plan


pinecone/core/openapi/shared/api_client.py
```diff
@@ -8,6 +8,24 @@
...
+import time
+import random
+
+def retry_api_call(
+    func, args=(), kwargs={}, retries=3, backoff=1, jitter=0.5
+):
+    attempts = 0
+    while attempts < retries:
+        try:
+            return func(*args, **kwargs)  # Attempt to call __call_api
```
harsha2010, line 20: I'm not a python expert, but isn;t there some library that provides a decorator to retry functions? seems like this should be a solved problem

pinecone/data/index.py
```diff
@@ -435,7 +470,58 @@ def query(
...
+            raise ValueError("At least one namespace must be specified")
+        if len(vector) == 0:
+            raise ValueError("Query vector must not be empty")
+
+        # The caller may only want the top_k=1 result across all queries,
+        # but we need to get at least 2 results from each query in order to
+        # aggregate them correctly. So we'll temporarily set topK to 2 for the
+        # subqueries, and then we'll take the topK=1 results from the aggregated
+        # results.
+        overall_topk = top_k if top_k is not None else 10
```
harsha2010, line 499: what is this magic number 10?

jhamon: It's just a default value. Could be anything, but the API requires a value to be passed.

pinecone/core/openapi/shared/api_client.py
jhamon: There's nothing in the standard library afaik, and I'm very hesitant to add third-party dependencies because of the overall dependency hell situation in python.

jhamon:
Pursuing this in a different branch #409

PR #410, opened 2024-10-31
[feat] PineconeGrpcFuture implements concurrent.futures.Future

jhamon:
## Problem

`GRPCIndex` has long had limited and poorly documented support for async operations via the futures interface of the [`grpc`](https://grpc.github.io/grpc/python/grpc.html#grpc.UnaryUnaryMultiCallable.future) library. I've recently been trying to implement `query_namespaces` using these futures, and discovered that unfortunately the grpc futures implementation is not compatible with the `concurrent.futures` package in the standard library. This makes them pretty much useless for anything at all complicated because the grpc library doesn't provide any utils for synchronization or waiting.

## Solution

A class called `PineconeGrpcFuture` was added in the past as a minimal wrapper around the [future](https://grpc.github.io/grpc/python/grpc.html#future-interfaces) that is emitted by `grpc`. These futures objects are used to represent asynchronous computation, and allow you to regisiter callbacks with `add_done_callback`. This is similar to calling `then()` on a javascript promise.

The original purpose of our `PineconeGrpcFuture` wrapper class seems to have been to implement some basic (very basic) error mapping, but for this diff I decided to extend the class to implement the `concurrent.futures.Future` interface. This allows the instances of `PineconeGrpcFuture` to be used with `concurrent.futures.as_completed` and `concurrent.futures.wait` utilities, which makes them dramatically more ergonomic to deal with. Unfortunately the grpc future is not compatible with the `concurrent.future` package out of the box.

For the unit tests of `PineconeGrpcFuture`, I had to make heavy use of mocking because all the various grpc classes are tightly coupled and can't be simply setup without performing actual network calls. This doesn't give me huge confidence it's actually working as expected, so as a sanity check I added some additional integration test coverage for `upsert`, `fetch`, and `delete` using `concurrent.futures.wait`.

## Type of Change

## Test Plan

Added unit and integration tests

### DIFF
--- a/pinecone/grpc/future.py
+++ b/pinecone/grpc/future.py
@@ -45,7 +45,7 @@ class PineconeGrpcFuture(ConcurrentFuture):
 
     def set_result(self, result):
         if self._result_transformer:
-            result = self.result_transformer(result)
+            result = self._result_transformer(result)
         return super().set_result(result)
 
     def cancel(self):
--- a/pinecone/grpc/future.py
+++ b/pinecone/grpc/future.py
@@ -5,21 +5,26 @@ from pinecone.exceptions.exceptions import PineconeException
 
 
 class PineconeGrpcFuture(ConcurrentFuture):
-    def __init__(self, grpc_future: GrpcFuture, timeout: Optional[int] = 10):
+    def __init__(
+        self, grpc_future: GrpcFuture, timeout: Optional[int] = 10, result_transformer=None
+    ):
         super().__init__()
         self._grpc_future = grpc_future
         self.default_timeout = timeout  # seconds
-
-        # Add callback to subscribe to updates from the gRPC future
-        self._grpc_future.add_done_callback(self._sync_state)
+        self.result_transformer = result_transformer
 
         # Sync initial state, in case the gRPC future is already done
         self._sync_state(self._grpc_future)
 
+        # Add callback to subscribe to updates from the gRPC future
+        self._grpc_future.add_done_callback(self._sync_state)
+
+    @property
+    def grpc_future(self):
+        return self._grpc_future
+
     def _sync_state(self, grpc_future):
-        # Sync the gRPC future completion to the wrapper future
         if self.done():
-            # Future already done, nothing to do
             return
 
         if grpc_future.cancelled():
@@ -34,6 +39,11 @@ class PineconeGrpcFuture(ConcurrentFuture):
                 self.set_exception(e)
         elif grpc_future.running():
             self.set_running_or_notify_cancel()
+
+    def set_result(self, result):
+        if self.result_transformer:
+            result = self.result_transformer(result)
+        return super().set_result(result)
 
     def cancel(self):
         self._grpc_future.cancel()
--- a/pinecone/grpc/future.py
+++ b/pinecone/grpc/future.py
@@ -1,34 +1,76 @@
-from grpc._channel import _MultiThreadedRendezvous
+from concurrent.futures import Future as ConcurrentFuture
+from typing import Optional
+from grpc import Future as GrpcFuture, RpcError
 from pinecone.exceptions.exceptions import PineconeException
 
 
-class PineconeGrpcFuture:
-    def __init__(self, delegate):
-        self._delegate = delegate
+class PineconeGrpcFuture(ConcurrentFuture):
+    def __init__(self, grpc_future: GrpcFuture, timeout: Optional[int] = 10):
+        super().__init__()
+        self._grpc_future = grpc_future
+        self.default_timeout = timeout  # seconds
+
+        # Add callback to subscribe to updates from the gRPC future
+        self._grpc_future.add_done_callback(self._sync_state)
+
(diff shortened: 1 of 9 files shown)


pinecone/grpc/future.py
```diff
@@ -1,34 +1,92 @@
...
+        else:
+            return self._default_timeout
+
+    def _wrap_rpc_exception(self, e):
+        if e._state and e._state.debug_error_string:
+            return PineconeException(e._state.debug_error_string)
+        else:
+            return PineconeException("Unknown GRPC error")
+
+    # def __repr__(self):
```
austin-denoble: Remove or uncomment? Assuming these were for debugging.

tests/unit_grpc/test_grpc_index_upsert.py
```diff
@@ -19,9 +19,27 @@ class MockUpsertDelegate:
     def __init__(self, upsert_response: UpsertResponse):
         self.response = upsert_response
 
-    def result(self, timeout):
+    def result(self, timeout=None):
```
aulorbe, line 22: Just OOC, did you set `timeout` to `None` here b/c you automatically set it to `5s` if it's `None` in  `future.py`?

aulorbe (approved):
lgtm, super clean and modular. Thanks for letting me take a look :)

austin-denoble (approved):
Really clean approach, LGTM. Thanks for all of the testing coverage too, very thorough.

And of course thanks for breaking these up into smaller prs.

pinecone/grpc/index_grpc.py
```diff
@@ -282,8 +282,12 @@ def delete(
...
 
     def fetch(
-        self, ids: Optional[List[str]], namespace: Optional[str] = None, **kwargs
-    ) -> FetchResponse:
+        self,
+        ids: Optional[List[str]],
+        namespace: Optional[str] = None,
+        async_req: Optional[bool] = False,
+        **kwargs,
+    ) -> Union[FetchResponse, PineconeGrpcFuture]:
```
haruska, line 290: Just for my understanding... `Union` is a way to have an "Either" return type? Or is it stating more the future ultimately has a `FetchResponse` shape once the promise is resolved?

Other question... is this ultimately preferred in Python over having two functions? (e.g. `fetch` and `fetch_async`)?

haruska (approved):

tests/unit_grpc/test_grpc_index_upsert.py
jhamon: This `MockUpsertDelegate` is a standin for the grpc future returned from the grpc runner. That object has this interface: https://grpc.github.io/grpc/python/_modules/grpc.html#Future.result

I adjusted the signature to match the `result` args in the interface.

pinecone/grpc/future.py
jhamon: Yeah, thanks for calling out. I'll clean these up.

pinecone/grpc/index_grpc.py
jhamon: > Just for my understanding... Union is a way to have an "Either" return type? Or is it stating more the future ultimately has a FetchResponse shape once the promise is resolved?

Yes, `Union[FetchResponse, PineconeGrpcFuture]` means you will get back either a `FetchResponse` or `PineconeGrpcFuture`.

> Other question... is this ultimately preferred in Python over having two functions? (e.g. fetch and fetch_async)?

I would strongly prefer to have methods with clear return types that don't require any casting or checking to know what you're getting, but if you look elsewhere in this class you'll see this is the general pattern used elsewhere in this class that has been present since before my time. It's something I would like to address in the future, but can't do it right now without creating a breaking change.

Eventually I would like to have something more like this: 

```python
# Sync interface, implemented for REST and GRPC
class SyncIndex(abc.ABC):
   def fetch(...) -> FetchResponse
class Index(SyncIndex):
   def fetch(...) -> FetchResponse
class GRPCIndex(SyncIndex):
   def fetch(...) -> FetchResponse

# Asyncio interface, implemented for REST and GRPC
class AsyncioIndex(abc.ABC):
   async def fetch(...) -> Awaitable[FetchResponse]
class AsyncioIndex(AsyncioIndex):
   async def fetch(...) -> Awaitable[FetchResponse]
class AsyncioIndexGRPC(AsyncioIndex):
   async def fetch(...) -> Awaitable[FetchResponse]


# Plus legacy async approaches, currently accessed with async_req=True kwarg
# These return types differ, so can't easily be grouped under one interface. Maybe with
# generics, which have some limited support in Python.

# Async, REST, using multiprocessing
class IndexAsync:
   def fetch(...) -> ApplyResult[FetchResponse]

# Async, GRPC, using futures
class GRPCIndexAsync:
   def fetch(...) -> PineconeGrpcFuture

```

tests/unit_grpc/test_grpc_index_upsert.py
aulorbe: ah thanks!

PR #420, opened 2024-12-03
[Refactor] Extract index request factory

jhamon:
## Problem

With my asyncio client in progress (not in this diff), I found myself needing to reuse a lot of the openapi request-building boilerplate.

## Solution

Rather than copy/paste and create a lot of duplication, I want to pull that logic out into a separate class for building request objects. This change should not break existing behavior in the Index client.

## Todo

This was pretty much just a mechanical extraction. I'd like to cleanup and standardize some stuff regarding `_check_type`, an openapi param that we're going to silly lengths to flip the default behavior on, in a follow-up diff.

## Type of Change

- [x] None of the above: Refactor

## Test Plan

Should still have tests passing

### DIFF
--- a/pinecone/data/import_error.py
+++ b/pinecone/data/import_error.py
@@ -1,0 +1,30 @@
+class IndexClientInstantiationError(Exception):
+    def __init__(self, index_args, index_kwargs):
+        formatted_args = ", ".join(map(repr, index_args))
+        formatted_kwargs = ", ".join(f"{key}={repr(value)}" for key, value in index_kwargs.items())
+        combined_args = ", ".join([a for a in [formatted_args, formatted_kwargs] if a.strip()])
+
+        self.message = f"""You are attempting to access the Index client directly from the pinecone module. The Index client must be instantiated through the parent Pinecone client instance so that it can inherit shared configurations such as API key.
+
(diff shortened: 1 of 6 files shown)


pinecone/data/import_error.py
```diff
@@ -0,0 +1,30 @@
+class IndexClientInstantiationError(Exception):
```
jhamon, line 1: Since I was already in a mode of pulling things out of `index.py`, I decided to move this to a different file as well.

pinecone/data/request_factory.py
```diff
@@ -0,0 +1,170 @@
...
+    return SparseValues(indices=sparse_values["indices"], values=sparse_values["values"])
+
+
+def non_openapi_kwargs(kwargs):
+    return {k: v for k, v in kwargs.items() if k not in OPENAPI_ENDPOINT_PARAMS}
+
+
+class IndexRequestFactory:
+    @staticmethod
+    def query_request(
```
austin-denoble, line 48: Good idea 👍 , I like encapsulating all of these helpers into a specific class.

austin-denoble (approved):
Nice one, this makes a lot of sense given all the boilerplate we need to deal with to construct these requests.

PR #421, opened 2024-12-06
[Bug] query_namespaces can handle single result

jhamon:
## Problem

In order to merge results across multiple queries, the SDK must know which similarity metric an index is using. For dotproduct and cosine indexes, a larger score is better while for euclidean a smaller score is better. Unfortunately the data plane API does not currently expose the metric type and a separate call to the control plane to find out seems undesirable from a resiliency and performance perspective.

As a workaround, in the initial implementation of `query_namespaces` the SDK would infer the similarity metric needed to merge results by seeing whether the scores of query results were ascending or descending. This worked well, but imposes an implicit limitation that there must be at least 2 results returned. 

We initially believed this would not be a problem but have since learned that applications using filtering can sometimes filter out all or most results. So an approach that has the user explicitly telling the SDK what similarity metric is being used is preferred to handle these edge cases with 1 or 0 results.

## Solution

- Add a required kwarg to `query_namespaces` to specify the index similarity metric. 
- Modify `QueryResultsAggregator` to use this similarity metric, and strip out code that was involved in inferring whether results were ascending or descending.
- Adjust integration tests to pass new metric kwarg. Except for adding the new kwarg, query_namespaces integration tests did not need to change which indicates the underlying behavior is still working as before.

## Type of Change

### DIFF
--- a/tests/integration/data/test_query_namespaces.py
+++ b/tests/integration/data/test_query_namespaces.py
@@ -1,15 +1,11 @@
 import pytest
 from ..helpers import random_string, poll_stats_for_namespace
-from pinecone.data.query_results_aggregator import (
-    QueryResultsAggregatorInvalidTopKError,
-    QueryResultsAggregregatorNotEnoughResultsError,
-)
 
 from pinecone import Vector
 
 
 class TestQueryNamespacesRest:
-    def test_query_namespaces(self, idx):
+    def test_query_namespaces(self, idx, metric):
         ns_prefix = random_string(5)
         ns1 = f"{ns_prefix}-ns1"
         ns2 = f"{ns_prefix}-ns2"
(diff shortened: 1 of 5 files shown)


tests/integration/data/test_query_namespaces.py
```diff
@@ -145,22 +147,7 @@ def test_query_namespaces(self, idx):
         assert len(results6.matches) == 0
         assert results6.usage.read_units > 0
 
-    def test_invalid_top_k(self, idx):
```
jhamon, line 148: Don't need this test anymore since top_k of 1 is now valid.

austin-denoble (approved):
LGTM, this seems like a reasonable way to do it. Plus, if the user has stored off their own index configuration details, or fetched them previously, they could just thread them through.

rohanshah18 (approved):
Tested locally for sorting and duplicates order. LGTM!

PR #418, opened 2024-11-15
[Feat] Helpful error when incorrectly instantiating Index client

jhamon:
## Problem

Sometimes users incorrectly instantiate the Index client like this:

```python
import pinecone

# Initialize the Index with the host
index = pinecone.Index(index_name, host=index_host)
```

Then they will later get an authentication exception when using it, since the `Index` class does not have the configuration values it needs when attempting to perform vector operations.

```python
ForbiddenException: (403) Reason: Forbidden HTTP response headers: HTTPHeaderDict({'Date': 'Wed, 13 Nov 2024 02:06:45 GMT', 'Content-Type': 'text/plain', 'Content-Length': '9', 'Connection': 'keep-alive', 'x-pinecone-auth-rejected-reason': 'Wrong API key', 'www-authenticate': 'Wrong API key', 'server': 'envoy'}) HTTP response body: Forbidden
```

## Solution

- Rename the `Index` implementation to `_Index` so that people will not accidentally interact with it.
- Add a new stub implementation for `Index` that throws an informative message.

Bonus:
- Move the Index client docstrings off the implementation class and into a related abstract base class. This wasn't strictly necessary, but I was feeling a bit overwhelmed by the size of the `index.py` file. I copied this approach from the grpc module. pdoc seems to still find and render the docs okay when doing this.

## Usage

The error message incorporates the args/kwargs the user was attempting to pass.

One positional arg
```python
>>> import pinecone
>>> i = pinecone.Index('my-index')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jhamon/workspace/pinecone-python-client/pinecone/data/index.py", line 113, in __init__
    raise IndexClientInstantiationError(args, kwargs)
pinecone.data.index.IndexClientInstantiationError: You are attempting to access the Index client directly from the pinecone module. The Index client must be instantiated through the parent Pinecone client instance so that it can inherit shared configurations such as API key.

    INCORRECT USAGE:
        ```
        import pinecone

        pc = pinecone.Pinecone(api_key='your-api-key')
        index = pinecone.Index('my-index')
        ```

    CORRECT USAGE:
        ```
        from pinecone import Pinecone

        pc = Pinecone(api_key='your-api-key')
        index = pc.Index('my-index')
        ```
```

Multiple positional args

```python
>>> i = pinecone.Index('my-index', 'https://my-index.blahblah.com')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #425, opened 2024-12-16
[Fix] Rename Index class to fix plugin installation

jhamon:
## Problem

By renaming the `Index` class to `_Index` in https://github.com/pinecone-io/pinecone-python-client/pull/418, I accidentally broke plugin installation.

## Solution

Rename the class back to `Index` and instead alias it to `_Index` through import statements. We still need to alias it within the `pinecone/data/__init__.py` module definition to avoid collision with the Index class designed to give users a more informative error when they incorrectly import the `Index` class from the `pinecone` module.

I went ahead and also cleaned up some stuff where we were unnecessarily exposing some internal openapi details in the public interface. Users should never realistically need to construct openapi request objects.

## Type of Change

### DIFF
--- a/pinecone/data/index.py
+++ b/pinecone/data/index.py
@@ -11,18 +11,10 @@ from pinecone.core.openapi.db_data.api.vector_operations_api import VectorOperat
 from pinecone.core.openapi.db_data import API_VERSION
 from pinecone.core.openapi.db_data.models import (
     FetchResponse,
-    QueryRequest,
     QueryResponse,
-    RpcStatus,
-    ScoredVector,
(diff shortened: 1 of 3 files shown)


jseldess (approved):
LGTM, though I'm not really qualified to say so :)

pinecone/data/index.py
```diff
@@ -45,32 +37,13 @@
 
 logger = logging.getLogger(__name__)
 
-__all__ = [
```
austin-denoble, line 48: Nice thinking getting these out of here. 👍

austin-denoble (approved):

PR #438, opened 2025-01-29
Migrate `create_index_for_model` from records plugin

jhamon:
## Problem

We need to migrate `create_index_for_model` functionality from the records plugin into the core of the SDK to provide improved UX around code completions and error handling
 
## Solution

- Copy & modify implementation of `create_index_for_model` and `IndexEmbed` from records plugin
- I did some moderately heavy refactoring of this code and also the existing `create_index` method to reduce the amount of duplication.
- Add records plugin to list of deprecated plugins that will halt startup with an exception
- Copy relevant tests from plugin and supplement with additional tests to ensure enum types and edge cases are handled properly even after refactoring
- Reuse recently-defined enums to make the method signature more self-documenting.

## Todo

- Data plane operations in the records plugin will be handled in a separate PR

## Usage

These would all be considered valid usage. Enums are available to help know what values are accepted, but you can type the literal strings if you prefer. This flexibility also keeps compatibility with existing usage of the plugin.

```python
from pinecone import (
    Pinecone, 
    EmbedModel, 
    CloudProvider, 
    AwsRegion, 
    IndexEmbed, 
    Metric
)

pc = Pinecone(api_key='key')

# All hand-crafted literals
pc.create_index_for_model(
    name='index-name',
    cloud='aws',
    region='us-east-1',
    embed={
        "model": "multilingual-e5-large", 
        "field_map": {"text": "my-sample-text"},
        "metric": "cosine"
    },
)

# All enum values
pc.create_index_for_model(
    name='index-name',
    cloud=CloudProvider.AWS,
    region=AwsRegion.US_EAST_1,
    embed=IndexEmbed(
        model=EmbedModel.Multilingual_E5_Large, 
        field_map={"text": "my-sample-text"},
        metric=Metric.COSINE
    ),
)

# Mixed literals and enums
pc.create_index_for_model(
    name='index-name',
    cloud='aws',
    region=AwsRegion.US_EAST_1,
    embed={
        "model": EmbedModel.Multilingual_E5_Large, 
        "field_map": {"text": "my-sample-text"},
        "metric": "cosine"
    },
)
```

## Type of Change


pinecone/models/index_embed.py
```diff
@@ -0,0 +1,68 @@
...
+    Optional.
+    """
+
+    def as_dict(self) -> Dict[str, Any]:
+        """
+        Returns the IndexEmbed as a dictionary.
+        """
+        return self.__dict__
+
+    def __init__(
```
austin-denoble, line 50: Thanks very much for improving this generally, I don't think I left things in the best state in the plugin itself.

pinecone/deprecated_plugins.py
```diff
@@ -8,10 +9,14 @@ def check_for_deprecated_plugins():
...
-            )
+            raise DeprecatedPluginError("pinecone-plugin-inference")
+    except ImportError:
+        pass
+
+    try:
+        from pinecone_plugins.records import __installables__  # type: ignore
+
+        if __installables__ is not None:
+            raise DeprecatedPluginError("pinecone-plugin-records")
```
austin-denoble, line 20: Really like your solution for this. 👍

austin-denoble (approved):
Great work, big thanks again for all the plugin porting. Loving the test expansion, and general UX improvement through enums, interface flexibility, etc. 🎉

PR #469, opened 2025-03-24
[KE] Add assistant description

avi1mizrahi:
## Problem

There were no info on assistant in the readme

## Solution

Now there is info on assistant in the readme

## Type of Change

## Test Plan

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,17 @@ response = index.search_records(
 )
 ```
 
+## Pinecone Assistant
+### Installing the Pinecone Assistant Python plugin
+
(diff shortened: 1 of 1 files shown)


nihohit:
it's better than nothing, but let's aim to have as much visibility as inference here.

avi1mizrahi:
> Thanks for adding a callout to the core python README.
> 
> The Assistant representation in the typescript client is fairly detailed if you're interested in modeling it around that: https://github.com/pinecone-io/pinecone-ts-client?tab=readme-ov-file#pinecone-assistant

Thanks Austin! Because we are currently have separated repos for the Python SDK and the plugin, this add another place we need to keep in sync. I hope soon we will be able to merge the plugin and add documentation only here.

PR #476, opened 2025-04-24
Remove urllib3 info from useragent string

jhamon:
## Problem

- Currently we send urllib3 version in the user agent string
- Testing with importtime shows importing urllib3 just to get the version extends the initial load time of the pinecone package by about 35 milliseconds.

## Solution

I confirmed we're not using this information in the header, so we should remove it to improve initialization performance.

## Test Plan

- Analysis with showed Loading this urllib3 to get the version was adding 35 milliseconds to the time needed to "from pinecone import Pinecone"
- Updated unit tests

### DIFF
--- a/pinecone/utils/user_agent.py
+++ b/pinecone/utils/user_agent.py
@@ -1,5 +1,3 @@
-import urllib3
-
 from .version import __version__
 from .constants import SOURCE_TAG
 import re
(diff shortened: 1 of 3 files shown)


jhamon:
I'm merging this with failures; failing tests are unrelated and I've filed a bug with the backend team.

PR #478, opened 2025-05-07
Fix: PineconeGrpcFuture blocks during construction

yorickvP:
## Problem

When using grpc with `async_req=True`, the construction of a `PineconeGrpcFuture` would call `_sync_state`, which would do a blocking call to `grpc_future.exception(...)`. This means that the async reqs were all blocking in practice.

## Solution

We can fix it by checking if the future is still running and not doing any blocking calls when it is.

## Type of Change

## Test Plan

How I tested it:

- Before: Run a long query in a loop and hit ctrl-c. Observe the traceback containing `PineconeGrpcFuture()`
- After: Run a long query in a loop and hit ctrl-c. Observe the traceback containing your `.result()` call instead.

### DIFF
--- a/pinecone/grpc/future.py
+++ b/pinecone/grpc/future.py
@@ -30,8 +30,9 @@ class PineconeGrpcFuture(ConcurrentFuture):
         if self.done():
             return
 
-        if grpc_future.running():
-            self.set_running_or_notify_cancel()
+        if grpc_future.running() and not self.running():
+            if not self.set_running_or_notify_cancel():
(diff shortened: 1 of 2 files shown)


jhamon:
This makes a lot of sense; I didn't realize `exception()` was blocking . Thanks for the patch.

jhamon:
seems like some of the unit tests are broken by this change. I'll need to do some follow-up before we can merge this in.

yorickvP:
I've fixed the unit test.

PR #490, opened 2025-05-16
Disable client-side validation of responses

jhamon:
## Problem

Sometimes unexpected values in API responses can cause unnecessary errors due to validation logic being applied. Fields labeled in the openapi as `enum` fields will error when unexpected values appear in the response. In general, we want the client to just display what the API returns without applying validation.

## Solution

Adjust the code generation to disable validation logic when instantiating model objects from API response.

## Type of Change

## Test Plan

I created a mock server script to do some manual testing of different responses with odd values in them and saw this works without erroring. For example, these responses no longer raise:
- New index status
- Index dimension > 20k
- Index name too long

### DIFF
--- a/pinecone/openapi_support/model_utils.py
+++ b/pinecone/openapi_support/model_utils.py
@@ -151,7 +151,10 @@ class OpenApiModel(object):
             # when listing indexes due to validation on the status field against
             # the allowed values in the enum.
             check_allowed_values(self.allowed_values, (name,), value)
-        if (name,) in self.validations:
+        if (name,) in self.validations and self._enforce_validations:
+            # Disabling validation on response makes the SDK
+            # less fragile if unexpected values are returned. In general,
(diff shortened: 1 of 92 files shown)


pinecone/openapi_support/model_utils.py
```diff
@@ -151,7 +151,10 @@ def set_attribute(self, name, value):
             # when listing indexes due to validation on the status field against
             # the allowed values in the enum.
             check_allowed_values(self.allowed_values, (name,), value)
-        if (name,) in self.validations:
+        if (name,) in self.validations and self._enforce_validations:
+            # Disabling validation on response makes the SDK
+            # less fragile if unexpected values are returned. In general,
+            # we want the SDK to display whatever is returned by the API.
```
jhamon, line 157: This is where the logic change actually comes into force.

PR #492, opened 2025-05-16
Add retry configuration for asyncio

jhamon:
## Problem

We want to use exponential backoff to retry failed requests made via PineconeAsyncio

## Solution

- Add `aiohttp-retry` dependency without the `asyncio` extras group
- Implement a JitterRetry class to calculate backoff intervals
    - The off-the-shelf JitteryRetry class has some odd behavior so i wanted to implement my own. This helps keep the behavior close to what we're doing for urllib3.
    - Intervals are roughly 0.1, 0.2, 0.4, 0.8 seconds (plus small jitter factor)
- Manual testing with test server in `scripts/test-server.py` and `scripts/test-async-retry.py`

## Type of Change

## Test Plan

Added some scripts for manual testing


scripts/test-async-retry.py
```diff
@@ -0,0 +1,16 @@
...
+import logging
+from pinecone import PineconeAsyncio
+
+dotenv.load_dotenv()
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+async def main():
+    async with PineconeAsyncio(host="http://localhost:8000") as pc:
```
austin-denoble, line 12: Was there also a `scripts/test-server.py` file?

austin-denoble (approved):

scripts/test-async-retry.py
jhamon: It's already in. It got committed in a previous diff for urllib3 retries.

austin-denoble: Thanks, was checking `main` not the branch 🤦‍♂️

PR #498, opened 2025-05-21
Fix inference autocomplete

jhamon:
## Problem

While implementing lazy loading and a package level .pyi file, we accidentally broke intellisense for some classes.

## Solution

- Add missing exceptions to pyi
- Remove type annotations on `pc.inference` that rely on values imported during `TYPE_CHECKING`. It seems like mypy and intellisense/autocomplete are able to infer these types without them being labeled explicitly. I think there is some sort of bug in the way autocomplete works because these type hints used with [TYPE_CHECKING](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) are valid and seem to work for typechecking purposes. 

## Type of Change

## Test Plan

Made a `7.0.1.dev1` dev build to confirm autocomplete behavior in different environments

### DIFF
--- a/pinecone/__init__.py
+++ b/pinecone/__init__.py
@@ -6,7 +6,21 @@ from .deprecated_plugins import check_for_deprecated_plugins as _check_for_depre
 from .deprecation_warnings import *
 from .pinecone import Pinecone
 from .pinecone_asyncio import PineconeAsyncio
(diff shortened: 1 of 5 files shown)


ssmith-pc (approved):
Works now!

PR #503, opened 2025-05-27
Fix errors related to missing readline dependency

jhamon:
## Problem

#488 

## Solution

This error was the result of the Cursor assistant putting some code into `pinecone/scripts/repl.py` that should only have gone into `scripts/repl.py` (and not been part of the package distribution). This wasn't caught in CI because `readline` is included in our dev dependencies, which are installed in CI for testing purposes.

## Type of Change

## Test Plan

Going to make a dev build and verify `list_models()` works in google colab (where dev dependencies are not present)


.github/workflows/pr.yaml
```diff
@@ -47,15 +47,19 @@ jobs:
     uses: './.github/workflows/testing-unit.yaml'
     secrets: inherit
 
-  integration-tests:
-      uses: './.github/workflows/testing-integration.yaml'
-      secrets: inherit
-      needs: unit-tests
+  # integration-tests:
```
austin-denoble, line 50: Do these commented blocks need to come out?

austin-denoble (approved):

PR #507, opened 2025-05-30
Add describe, delete, and list namespaces (REST)

rohanshah18:
## Problem

Add support for describe, delete, and and list namespaces for index and async index classes.

## Solution

Following methods were added to the index and async index classes:
1. describe_namespace(self, namespace: str, **kwargs) -> "NamespaceDescription"
2. delete_namespace(self, namespace: str, **kwargs) -> Dict[str, Any]
3. list_namespaces:
        - list_namespaces(self, **kwargs) -> Iterator[ListNamespacesResponse] (for index.py)
        - list_namespaces(self, **kwargs) -> AsyncIterator[ListNamespacesResponse] (for index_asyncio.py)
4. list_namespaces_paginated(
        self, limit: Optional[int] = None, pagination_token: Optional[str] = None, **kwargs
    ) -> ListNamespacesResponse

## Type of Change

## Test Plan

Added integration tests for both index and async index classes


tests/integration/helpers/helpers.py
```diff
@@ -280,11 +280,7 @@ async def asyncio_wait_until(
 
 
 def default_create_index_params(request, run_id):
-    github_actor = os.getenv("GITHUB_ACTOR", None)
-    user = os.getenv("USER", None)
-    index_owner = github_actor or user or "unknown"
-
-    index_name = f"{index_owner}-{str(uuid.uuid4())}"
```
rohanshah18, line 287: removed {index_owner} because with the current pattern and my user-name, the length(index name) > 45 chars.

tests/integration/data/test_namespace.py
```diff
@@ -0,0 +1,159 @@
...
+                index.delete_namespace(namespace.name)
+            except Exception as e:
+                print(f"Error deleting namespace {namespace.name}: {e}")
+
+        # Wait for deletions to complete
+        time.sleep(5)
+    except Exception as e:
+        print(f"Error in delete_all_namespaces: {e}")
+
+@pytest.mark.skipif(
```
rohanshah18, line 46: Will enable this once grpc endpoints for namespaces are added

pinecone/db_data/resources/sync/namespace_request_factory.py
```diff
@@ -0,0 +1,29 @@
...
+
+class DescribeNamespaceArgs(TypedDict, total=False):
+    namespace: str
+
+
+class DeleteNamespaceArgs(TypedDict, total=False):
+    namespace: str
+
+
+class NamespaceRequestFactory:
```
rohanshah18, line 14: Followed the same pattern as Imports

pinecone/db_data/resources/sync/namespace.py
```diff
@@ -0,0 +1,120 @@
...
+class NamespaceResource(PluginAware):
+    def __init__(
+        self,
+        api_client,
+        config,
+        openapi_config,
+        pool_threads: int,
+    ) -> None:
+        self.config = config
+        """ :meta private: """
```
rohanshah18, line 26: Updated """@private""" with """ :meta private: """ based on this [pr](https://github.com/pinecone-io/pinecone-python-client/pull/508)

pinecone/db_data/resources/sync/namespace_request_factory.py
```diff
@@ -0,0 +1,29 @@
...
+    @staticmethod
+    def describe_namespace_args(namespace: str) -> DescribeNamespaceArgs:
+        return {"namespace": namespace}
```
austin-denoble: Does this need the same `if isinstance(namespace, int):` check as below?

tests/integration/data/test_namespace.py
austin-denoble, line 46: Sort of nitpicky, but I think instead of doing a `skipif` check and worrying about an env variable, you could just skip always and then update in the gRPC pr:

```suggestion
@pytest.mark.skip(reason="enable when gRPC implemented")
```

tests/integration/data_asyncio/test_namespace_asyncio.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # Verify namespace is deleted
+        assert not await verify_namespace_exists(asyncio_idx, test_namespace)
+
+    @pytest.mark.asyncio
+    async def test_list_namespaces(self, index_host):
+        """Test listing namespaces"""
+        asyncio_idx = build_asyncioindex_client(index_host)
+        # Create multiple test namespaces
+        test_namespaces = ["test_list_1_async", "test_list_2_async", "test_list_3_async"]
```
austin-denoble, line 90: Might be worth a fixture in `conftest`? It's been a while since I've looked at how the tests are setup in this repo, but it could be worth trying to have a shared fixture per-session that works with namespaces.

tests/integration/helpers/helpers.py
austin-denoble: Seems like the ownership is helpful in the index name, could we just slice off a bunch of the uuid instead?

`index_name = f"{index_owner}-{str(uuid.uuid4())[:5]}"`

pinecone/db_data/index.py
```diff
@@ -605,3 +628,21 @@ def cancel_import(self, id: str):
...
+    @validate_and_convert_errors
+    def describe_namespace(self, namespace: str) -> "NamespaceDescription":
+        return self.namespace.describe(namespace=namespace)
+
+    @validate_and_convert_errors
+    def delete_namespace(self, namespace: str) -> Dict[str, Any]:
+        return self.namespace.delete(namespace=namespace)
+
+    @validate_and_convert_errors
+    def list_namespaces(self, **kwargs) -> Iterator[ListNamespacesResponse]:
```
austin-denoble: Should this just be `**kwargs` like this?

pinecone/db_data/resources/asyncio/namespace_asyncio.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        Args:
+            namespace (str): The namespace to delete
+
+        Delete a namespace from an index.
+        """
+        args = NamespaceRequestFactory.delete_namespace_args(namespace=namespace)
+        return await self.__namespace_operations_api.delete_namespace(**args)
+
+    async def list(self, **kwargs) -> AsyncIterator[ListNamespacesResponse]:
```
austin-denoble: I asked about the function signature being `**kwargs` elsewhere, is there a reason it can't be the same as `list_paginated`? I might be missing something in terms of returning the `AsyncIterator`.

austin-denoble (commented):
LGTM overall, left some comments and a few questions. Nice work.

pinecone/db_data/index.py
rohanshah18: No, self is needed for accessing NamespaceResource instance. Notice line#642 calling self.namespace.list().
Now you may ask why do we need self in NamespaceResource's list() - that's because it further calls self.list_paginated().

pinecone/db_data/resources/asyncio/namespace_asyncio.py
rohanshah18: Regardless of Async or sync, list() has **kwargs because it allows users to pass in limit if needed. The main difference between list() and list_paginated is that the list() handles pagination automatically.

pinecone/db_data/resources/sync/namespace_request_factory.py
rohanshah18: I actually forgot to remove this check but @jhamon do you think I should include type check? If I understand the client correctly, the api will be called with int and the server will then throw an error.

tests/integration/helpers/helpers.py
rohanshah18: @jhamon and I agreed to remove index-owner so do you see any use-case for having ownership in the index-name? Because all of the indexes are deleted as a part of post-step of the integration test.

jhamon: I told Rohan this was fine. The name thing is vestigial now that all test runs execute in the context of a fresh project. I'm going to be making more changes to this soon so that index names are fully deterministic and repeatable.

pinecone/db_data/resources/sync/namespace.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        Args:
+            namespace (str): The namespace to delete
+
+        Delete a namespace from an index.
+        """
+        args = NamespaceRequestFactory.delete_namespace_args(namespace=namespace)
+        return self.__namespace_operations_api.delete_namespace(**args)
+
+    def list(self, **kwargs) -> Iterator[ListNamespacesResponse]:
```
jhamon: I would adjust the method signature to spell out the parameters you described in the docstrings. This will better support code completion and type hints in peoples code editors because it gives a better idea of what arguments are expected.

```suggestion
    def list(self, limit: Optional[int] = None, pagination_token: Optional[str] = None, **kwargs) -> Iterator[ListNamespacesResponse]:
```

To use `Optional` type hint, import it like `from typing import Optional`. If these aren't passed, they will get a default value of `None` (basically the same as null in python)

As for `**kwargs`, we use **kwargs to gather up all other keyword arguments passed into a dictionary. For example `list(limit=10, foo='bar')` would put `'foo'` into the `kwargs` dictionary but not `limit`. We typically only rely on this for secret / undocumented behaviors, or when we need some way to pass arguments through to somewhere else without creating a tight coupling to the list() signature. 

A common example of how this might be used, in theory anyway, is to pass kwargs to the underlying openapi API method. But we don't want openapi junk leaking out into public view and polluting our method interface, so we don't want to declare all those properties on every method we implement.

pinecone/db_data/resources/sync/namespace.py
```diff
@@ -0,0 +1,120 @@
...
+                to fetch the next page of results. [optional]
+
+        Returns:
+            Returns a generator that yields each namespace. It automatically handles pagination tokens on your behalf so you can
+            easily iterate over all results. The `list` method accepts all of the same arguments as list_paginated
+
+        ```python
+        for namespace in index.list_namespaces():
+            print(namespace)
+        ```
```
jhamon: Docstrings now need to use RST syntax for sphinx to parse them correctly. Instead of triple bacticks, you need to use the `.. code-block:: python` directive. The example code needs to be indented one line from the indent level of the directive. And (very important) you have to have a blank line between the directive start and the first line of your code, otherwise nothing will be rendered. You can search `code-block` in the repo to find many examples

```suggestion
       .. code-block:: python
        
          for namespace in index.list_namespaces():
              print(namespace)
```

pinecone/db_data/resources/sync/namespace.py
```diff
@@ -0,0 +1,120 @@
...
+    def list(self, **kwargs) -> Iterator[ListNamespacesResponse]:
+        """
+        Args:
+            limit (Optional[int]): The maximum number of namespaces to fetch in each network call. If unspecified, the server will use a default value. [optional]
+            pagination_token (Optional[str]): When there are multiple pages of results, a pagination token is returned in the response. The token can be used
+                to fetch the next page of results. [optional]
+
+        Returns:
+            Returns a generator that yields each namespace. It automatically handles pagination tokens on your behalf so you can
+            easily iterate over all results. The `list` method accepts all of the same arguments as list_paginated
```
jhamon:
```suggestion
            easily iterate over all results. The ``list`` method accepts all of the same arguments as list_paginated
```

RST wants us to double-backtick code terms.

Here's a guide https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

pinecone/db_data/resources/sync/namespace.py
```diff
@@ -0,0 +1,120 @@
...
+        ```python
+        for namespace in index.list_namespaces():
+            print(namespace)
+        ```
+
+        You can convert the generator into a list by wrapping the generator in a call to the built-in `list` function:
+
+        ```python
+        namespaces = list(index.list_namespaces())
+        ```
```
jhamon: Same issue here with `.. code-block:: python` directive

pinecone/db_data/resources/sync/namespace.py
```diff
@@ -0,0 +1,120 @@
...
+
+        List all namespaces in an index with pagination support. The response includes pagination information if there are more results available.
+
+        Consider using the `list` method to avoid having to handle pagination tokens manually.
+
+        Examples:
+            >>> results = index.list_paginated(limit=5)
+            >>> results.pagination.next
+            eyJza2lwX3Bhc3QiOiI5OTMiLCJwcmVmaXgiOiI5OSJ9
+            >>> next_results = index.list_paginated(limit=5, pagination_token=results.pagination.next)
```
jhamon: These also need to be adjusted to use the `.. code-block:: python` directive.

pinecone/db_data/resources/sync/namespace_request_factory.py
jhamon: Yeah, I would say either do this type coercion everywhere or nowhere; we don't want to have inconsistent behavior from one method to the next. 

Overall, instead of trying to be clever and coerce types, I think it's easier to just raise an exception if a non-string value is passed since that's probably a mistake the user has made rather than an intentional desire to set the namespace to an int or other non-string value.

E.g. I would do something like this

```python
if not isinstance(namespace, str):
    raise ValueError('namespace must be string')
```

pinecone/db_data/resources/asyncio/namespace_asyncio.py
```diff
@@ -0,0 +1,104 @@
...
+    ListNamespacesResponse,
+    NamespaceDescription,
+)
+
+from pinecone.utils import install_json_repr_override
+
+from ..sync.namespace_request_factory import NamespaceRequestFactory
+
+for m in [ListNamespacesResponse, NamespaceDescription]:
+    install_json_repr_override(m)
```
jhamon, line 14: I would do some manual testing with `poetry run repl` to make sure the output looks good for these methods. The goal of these overrides is to massage the output for notebook-style interactive use, but since it's relying on jsonifying other objects we need to make sure that's working well each time we add something new since the json dump may not understand all object types that exist.

pinecone/db_data/resources/asyncio/namespace_asyncio.py
jhamon: You may still want to pass `pagination_token` to `list()` if you want to resume iteration from somewhere other than the first page of results. I would add `limit` and `pagination_token` as explicit, but optional, arguments here instead of only relying on `**kwargs`.

tests/integration/data/test_namespace.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def delete_all_namespaces(index):
+    """Helper function to delete all namespaces in an index"""
+    try:
+        # Get all namespaces
+        namespaces = list(index.list_namespaces())
+
+        # Delete each namespace
+        for namespace in namespaces:
```
jhamon, line 35: When you call `list()`, that kind of defeats the purpose of returning an iterator. In this case it's fine, but if you had millions of namespaces it would try to fetch all the pages and return them in a single list which would be bad.


```suggestion
        # Delete each namespace
        for namespace in index.list_namespaces():
```

tests/integration/data/test_namespace.py
jhamon: I think Rohan's doing the right thing in this instance, actually. These tests are supposed to eventually run for grpc and for rest, although for now only REST is implemented. Skipped it totally would mean no testing of the REST code he has in this PR.

tests/integration/data/test_namespace.py
```diff
@@ -0,0 +1,159 @@
...
+            try:
+                index.delete_namespace(namespace.name)
+            except Exception as e:
+                print(f"Error deleting namespace {namespace.name}: {e}")
+
+        # Wait for deletions to complete
+        time.sleep(5)
+    except Exception as e:
+        print(f"Error in delete_all_namespaces: {e}")
```
jhamon: I would add this to the top of your test file:

```python
import logging

```

Then use it like `logger.error(f"Error in delete_all_namespaces: {e}")` instead of `print`. This way the output will have more context about where it came from. You don't see it in all tests, but lately I've been logging more and more info from tests and test setup/cleanup because the time needed to retry things is high and I want to minimize the amount of retrying. just to get information about what went wrong.

tests/integration/data/test_namespace.py
```diff
@@ -0,0 +1,159 @@
+import os
+import time
+
+import pytest
+
+from pinecone.core.openapi.db_data.models import NamespaceDescription
```
jhamon: You probably need to make some changes in `pinecone/__init__.py` and `pinecone/__init__.pyi` in order to get this to work, but you should be able to write `from pinecone import NamespaceDescription`. Users may want to access these objects for typing purposes, but we don't want them having to drill down into the internals to do it.

tests/integration/data/test_namespace.py
```diff
@@ -0,0 +1,159 @@
...
+            assert len(namespaces) == len(test_namespaces)
+            namespace_names = [ns.name for ns in namespaces]
+            for test_ns in test_namespaces:
+                assert test_ns in namespace_names
+
+            # Verify each namespace has correct structure
+            for ns in namespaces:
+                assert isinstance(ns, NamespaceDescription)
+                assert hasattr(ns, 'name')
+                assert hasattr(ns, 'vector_count')
```
jhamon, line 104: These are a fine start, but I usually would write assertions like this

```python
assert ns.name is not None
assert isinstance(ns.name, str)
assert ns.vector_count is not None
assert isinstance(ns.vector_count, int)

# Veriify object supports dict-style access
assert ns.name == ns['name']
assert ns.vector_count == ns['vector_count']

# Verify object supports get-style access
assert ns.get('name', None) == ns.name
```

pinecone/db_data/resources/sync/namespace.py
rohanshah18: I have followed the same pattern as list vectors and list imports, where we only pass all arguments as optional in the list_paginated method. 
list: https://github.com/pinecone-io/pinecone-python-client/blob/main/pinecone/db_data/index.py#L472
list imports: https://github.com/pinecone-io/pinecone-python-client/blob/main/pinecone/db_data/index.py#L516

pinecone/db_data/index.py
austin-denoble: I understand why self is needed, I was asking why we aren't using named args here rather than just a single kwargs, but maybe I was misunderstanding something.

PR #510, opened 2025-06-11
Fix grpc timeouts when async_req=True

jhamon:
## Problem

When users pass `async_req=True` and a `timeout` value to GRPC methods, the timeout value was not being honored; instead, a default timeout of 5 seconds was being imposed by the `PineconeGrpcFuture`. 

## Solution

The future object returned from the `grpc` package

## Type of Change

## Test Plan


pinecone/grpc/base.py
```diff
@@ -29,7 +29,15 @@ def __init__(
         _endpoint_override: Optional[str] = None,
     ):
         self.config = config
-        self.grpc_client_config = grpc_config or GRPCClientConfig()
+        # If grpc_config is passed, use it. Otherwise, build a new one with
+        # default values and passing in the ssl_verify value from the config.
+        if self.config.ssl_verify is None:
+            default_grpc_config = GRPCClientConfig()
+        else:
+            default_grpc_config = GRPCClientConfig(secure=self.config.ssl_verify)
```
jhamon, line 37: This change allows me to test against a local grpc server implementation with `PineconeGrpc(ssl_veriify=False)` setting getting carried through all the way to the index class.

pinecone/grpc/utils.py
```diff
@@ -63,9 +64,28 @@ def parse_usage(usage: dict):
...
+    upserted_count = json_response.get("upsertedCount", 0)
+    return UpsertResponse(upserted_count=int(upserted_count))
+
+
+def parse_update_response(response: Union[dict, Message], _check_type: bool = False):
+    return {}
+
+
+def parse_delete_response(response: Union[dict, Message], _check_type: bool = False):
+    return {}
```
jhamon, line 78: These two requests don't actually return anything, so there's nothing to do here.

tests/integration/data_grpc_futures/test_delete_future.py
```diff
@@ -1,14 +1,10 @@
...
-import pytest
 from pinecone import Vector
 from ..helpers import poll_stats_for_namespace, random_string
 import logging
+import time
 
 logger = logging.getLogger(__name__)
 
-if os.environ.get("USE_GRPC") == "true":
-    from pinecone.grpc import GRPCDeleteResponse
```
jhamon, line 10: Since I moved these tests into a new folder `data_grpc_futures/` that is only going to be run with grpc, I don't need as much complexity in the test setup.

tests/integration/data_grpc_futures/test_timeouts.py
```diff
@@ -0,0 +1,391 @@
...
+        result = query_results.result()
+        assert result is not None
+        assert isinstance(result, QueryResponse)
+        assert result.usage.read_units == 1
+        assert result.matches[0].id == "1"
+
+
+@pytest.mark.usefixtures("grpc_server")
+class TestGrpcAsyncTimeouts_QueryByVector:
+    def test_query_by_vector_with_timeout(self, local_idx: GRPCIndex):
```
rohanshah18: nit: test_query_by_vector_with_timeout -> test_query_by_vector_with_timeout_exceeded to match rest of the tests

tests/integration/data_grpc_futures/test_timeouts.py
```diff
@@ -0,0 +1,391 @@
...
+        )
+
+        assert len(done) == 1
+        assert len(not_done) == 0
+
+        result = update_results.result()
+        assert result is not None
+        assert result == {}
+
+
```
rohanshah18, line 288: Did you mean to leave out test_update_with_default_timeout test?

pinecone/grpc/base.py
```diff
@@ -29,7 +29,15 @@ def __init__(
         _endpoint_override: Optional[str] = None,
     ):
         self.config = config
-        self.grpc_client_config = grpc_config or GRPCClientConfig()
+        # If grpc_config is passed, use it. Otherwise, build a new one with
+        # default values and passing in the ssl_verify value from the config.
+        if self.config.ssl_verify is None:
```
rohanshah18, line 34: I'm a little unclear on why we are checking on ssl_verify property for setting default_grpc_config = GRPCClientConfig()

rohanshah18 (approved):
Really nice work on adding the tests for timeouts! Left a few small comments and an additional comment that the test_update_future.py file is empty, so not sure if you want to delete the file or add new tests. LGTM!!!

tests/integration/data_grpc_futures/test_query_future.py
```diff
@@ -0,0 +1,350 @@
...
+        assert find_by_id(query_result.matches, "6") is not None
+
+    @pytest.mark.skip(reason="Seems like a bug in the server")
+    def test_query_by_id_with_filter_nin(self, idx, query_namespace, use_nondefault_namespace):
+        target_namespace = query_namespace if use_nondefault_namespace else ""
+
+        # Vector(id='4', values=embedding_values(2), metadata={'genre': 'action', 'runtime': 120 }),
+        # Vector(id='5', values=embedding_values(2), metadata={'genre': 'comedy', 'runtime': 90 }),
+        # Vector(id='6', values=embedding_values(2), metadata={'genre': 'romance', 'runtime': 240 })
+        query_result = idx.query(
```
rohanshah18, line 304: This test is being skipped so no need to update it right now but if we do decide to remove the skip decorator, we need to add async_req=true.

pinecone/grpc/base.py
jhamon: The `config` object here has an optional boolean field called `ssl_verify` that can have three values: `None` (by default, if unset by the user), `True`, or `False`. The user sets these by passing a value to the `Pinecone` or `PineconeGRPC` constructor like this: `pc = PineconeGRPC(ssl_verify=False)`

Separately, GRPC configurations can be set at the index level with `GRPCClientConfig` by passing `pc.Index(name='my-index', grpc_config=grpc_config)`. `GRPCClientConfig` is implemented as a `NamedTuple`, an immutable data structure. This object controls the SSL behavior any many other things within the grpc code. For SSL verification, the relevant property in this object is called `secure`. I discovered in testing that unlike `ssl_verify`, `secure` is a true boolean (True/False only) and if you set None you will get errors.

For configuring SSL verify behavior in GRPC, there are three scenarios we need to consider:
- The user never set anything in particular regarding SSL verification; they want the default behavior everywhere. By the time this constructor is executing, `config.ssl_verify` will be `None` so we use the `GRPCClientConfig()` with no modifications to configure the GRPC channel. This has SSL verification on by default.
- The user passed `PineconeGRPC(ssl_verify=False)` but did not specify a `grpc_config` object for the index client; in that scenario, we want to thread that setting through even to the index level and override the default grpc behavior for ssl verification.
- The user explicitly passed `grpc_config` to the index constructor with `pc.Index(name='foo', grpc_config=grppc_config)`. In this scenario, we use the configuration the user has provided. 

The lesson here is that we should endeavor to have only one way to configure things if at all possible. Merging multiple config sources, resolution order, etc gets complicated in a hurry.

PR #515, opened 2025-06-16
Add `embed` to Index `configure` calls

austin-denoble:
## Problem
`embed` was never exposed as an argument for calling `configure` on `IndexResource`.

## Solution
- Add a new simple `ConfigureIndexEmbed(TypedDict)` class for representing the argument dictionary shape. I went with this because it aligned with the existing `CreateIndexForModelEmbedTypedDict`, but I'm not sure if this is best practice in the repo at this point. Maybe a class would be better.
- Update factory, sync, and async resources to pass through `embed` on `configure` calls.
- Update legacy `Pinecone.configure_index` method to support `embed`.
- Add integration tests to serverless resources to validate converting an existing serverless index to an integrated index using `configure` or `configure_index`.

## Type of Change

## Test Plan
New integration tests added. You can pull this branch down and use poetry to run a repl and quickly evaluate things yourself:

```python
poetry run repl

>>> pc = Pinecone(api_key="YOUR_API_KEY")

>>> pc.create_index(name="test-int-inf-convert", dimension=1024, metric="cosine", spec={"serverless": {"cloud": "aws", "region": "us-east-1"}})
{
    "name": "test-int-inf-convert",
    "metric": "cosine",
    "host": "test-int-inf-convert-bt8x3su.svc.preprod-aws-0.pinecone.io",
    "spec": {
        "serverless": {
            "cloud": "aws",
            "region": "us-east-1"
        }
    },
    "status": {
        "ready": true,
        "state": "Ready"
    },
    "vector_type": "dense",
    "dimension": 1024,
    "deletion_protection": "disabled",
    "tags": null
}

>>> pc.db.index.configure(name="test-int-inf-convert", embed={"model": "multilingual-e5-large", "field_map":{"text": "chunk_text"}})
>>> pc.db.index.describe(name="test-int-inf-convert")
{
    "name": "test-int-inf-convert",
    "metric": "cosine",
    "host": "test-int-inf-convert-bt8x3su.svc.preprod-aws-0.pinecone.io",
    "spec": {
        "serverless": {
            "cloud": "aws",
            "region": "us-east-1"
        }
    },
    "status": {
        "ready": true,
        "state": "Ready"
    },
    "vector_type": "dense",
    "dimension": 1024,
    "deletion_protection": "disabled",
    "tags": null,
    "embed": {
        "model": "multilingual-e5-large",
        "field_map": {
            "text": "chunk_text"
        },
        "dimension": 1024,
        "metric": "cosine",
        "write_parameters": {
            "input_type": "passage",
            "truncate": "END"
        },
        "read_parameters": {
            "input_type": "query",
            "truncate": "END"
        },
        "vector_type": "dense"
    }
}

### repeat with async resources / pc.configure_index()
```

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210417294961252


pinecone/__init__.py
```diff
@@ -98,6 +98,11 @@
     "RestoreJobList": ("pinecone.db_control.models", "RestoreJobList"),
     "BackupModel": ("pinecone.db_control.models", "BackupModel"),
     "BackupList": ("pinecone.db_control.models", "BackupList"),
+    "ConfigureIndexEmbed": ("pinecone.db_control.types", "ConfigureIndexEmbed"),
```
rohanshah18, line 101: what's the reason for adding ConfigureIndexEmbed and CreateIndexForModelEmbedTypedDict here?

rohanshah18 (approved):
LGTM! Nice work. Also it makes sense to follow `CreateIndexForModelEmbed` approach.

pinecone/__init__.py
austin-denoble: Adding the classes to `_db_control_lazy_imports`, which we seem to do for all of our custom models and types. It seemed like `CreateIndexForModelEmbedTypedDict` was also not included here, so I added it.

This is to allow lazy loaded imports to be exported from the top level of the package here: https://github.com/pinecone-io/pinecone-python-client/blob/dfd01253344453540fae372beda354d589124709/pinecone/__init__.py#L145

rohanshah18: yeah I was wondering why we added types here but here's the reason why:
https://github.com/pinecone-io/pinecone-python-client/pull/507#discussion_r2140683425

PR #517, opened 2025-06-26
Add support for list, describe, and delete namespaces in grpc

rohanshah18:
## Problem

Support for REST sync and async namespaces was added in a previous PR, so this PR adds support for gRPC only.

## Solution

Add support for list, describe, and delete namespaces in grpc.

```python
from pinecone import Pinecone

# Initialize Pinecone client
pc = Pinecone(api_key='your-api-key')

# Get index
index = pc.Index('your-index-name')

# Example: List all namespaces (automatic pagination)
print("All namespaces:")
for namespace in index.list_namespaces():
    print(f"{namespace.name}: {namespace.record_count} records")

# Example: List namespaces with pagination control
results = index.list_namespaces_paginated(limit=10)
print(f"First 10 namespaces:")
for namespace in results.namespaces:
    print(f"{namespace.name}: {namespace.record_count} records")

# Example: Describe a specific namespace
namespace_info = index.describe_namespace(namespace='my-namespace')
print(f"Namespace '{namespace_info.name}' has {namespace_info.record_count} records")

# Example: Delete a namespace (irreversible!)
index.delete_namespace(namespace='namespace-to-delete')
```

## Type of Change

## Test Plan

Added integration and unit tests


pinecone/core/grpc/protos/db_data_2025_04_pb2.py
```diff
@@ -0,0 +1,146 @@
+# -*- coding: utf-8 -*-
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
+# NO CHECKED-IN PROTOBUF GENCODE
+# source: db_data_2025-04.proto
```
austin-denoble, line 4: Interesting, I didn't realize the proto files weren't also generated for the v7 release initially. Do the `2025_01` files need to be removed as well? Currently, 2025-01 and 2025-04 are present and I don't think that's correct:
Screenshot 2025-06-27 at 11 07 01 AM

rohanshah18: Yeah, I don't have much context here on whether we delete the older versions or not, but it definitely feels weird not deleting 2025_01. In the future, I'd like to add the following steps to `build-grpc.sh`:

1. Before code generation: Delete old proto files.
2. After code generation: Either
     - Update all `yyyy-mm` references in the codebase so we don’t have to manually edit the imports, or
     - Rename the proto files to remove `yyyy-mm`, so no edits are required in the codebase.

PR #525, opened 2025-10-13
Update protobuf to 5.29.5 to address security vulnerability

rohanshah18:
## Problem

The Pinecone Python client is currently using protobuf version `^5.29`, which includes vulnerable versions that are affected by [GHSA-8qvm-5x2c-j2w7](https://github.com/advisories/GHSA-8qvm-5x2c-j2w7). This vulnerability involves uncontrolled recursion in Protobuf's pure-Python backend, which could lead to Denial of Service (DoS) attacks.

## Solution

Updated the protobuf dependency constraint from `^5.29` to `^5.29.5` to ensure we're using the patched version that addresses this security vulnerability. The changes include:

- Updated `pyproject.toml`: Changed protobuf version constraint from `^5.29` to `^5.29.5`
- Updated `testing-dependency-grpc.yaml`: Updated protobuf version from `5.29.1` to `5.29.5` in all three dependency testing matrix configurations
- Verified that `poetry.lock` already contains protobuf 5.29.5, so no additional lock file updates were needed

This is a patch version update, so no breaking changes are expected. The protobuf dependency is optional and only installed when the `grpc` extra is requested.

**Note:** This is a security patch release to address the immediate vulnerability for existing users. A future release will include a comprehensive update to protobuf 6.x, which may include breaking changes and will require more extensive testing and migration planning.

## Type of Change

## Test Plan

- Verified protobuf 5.29.5 is already installed and working
- Updated CI/CD pipeline to test with the new version
- No breaking changes expected as this is a patch version update


rohanshah18:
Admin api tests are failing but they are safe to ignore because of some internal changes to org which resulted in max_pods = 20 instead 0.

PR #565, opened 2025-12-16
Expand protobuf version constraint to include v6.x

rohanshah18:
## Problem

The current protobuf dependency constraint (`>=5.29.5,<6.0.0`) blocks users from using protobuf 6.x. Some users may want or need to use protobuf 6.x in their environments, and the current upper bound prevents this.

## Solution

Updated the protobuf version constraint from `<6.0.0` to `<7.0.0` in `pyproject.toml`, allowing both protobuf 5.x and 6.x.

According to the [Protocol Buffers Cross-Version Runtime Guarantee](https://protobuf.dev/support/cross-version-runtime-guarantee/), code generated for major version V is supported by runtimes of versions V and V+1. Our proto files were generated with protobuf 5.x, so they should be compatible with 6.x runtimes.

**Verified locally:** Successfully imported proto modules and ran all gRPC unit tests with protobuf 6.33.2 installed.

**Security note:** CVE-2025-4565 is fixed in protobuf 5.29.5 and 6.31.1. The existing `>=5.29.5` lower bound ensures 5.x users get patched versions.

## Type of Change

## Test Plan

1. Ran all gRPC unit tests with protobuf 6.33.2 installed
2. CI will run the full test suite

### DIFF
--- a/.github/workflows/publish-to-pypi.yaml
+++ b/.github/workflows/publish-to-pypi.yaml
@@ -83,7 +83,7 @@ jobs:
 
       - name: Bump pyproject.toml version
         run: |
-          python -c "import re; content = open('pyproject.toml').read(); content = re.sub(r'version = \"[^\"]+\"', 'version = \"${{ steps.bump.outputs.version }}\"', content); open('pyproject.toml', 'w').write(content)"
+          python -c "import re; content = open('pyproject.toml').read(); content = re.sub(r'^version = \"[^\"]+\"', 'version = \"${{ steps.bump.outputs.version }}\"', content, flags=re.MULTILINE); open('pyproject.toml', 'w').write(content)"
 
       - name: Build Python client
         run: make package
--- a/uv.lock
+++ b/uv.lock
@@ -1580,7 +1580,7 @@ wheels = [
 
 [[package]]
 name = "pinecone"
-version = "7.3.0"
+version = "8.0.0"
 source = { editable = "." }
 dependencies = [
     { name = "certifi" },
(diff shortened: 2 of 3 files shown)


uv.lock
```diff
@@ -1580,7 +1580,7 @@ wheels = [
 
 [[package]]
 name = "pinecone"
-version = "7.3.0"
+version = "8.0.0"
```
rohanshah18, line 1583: Not sure why version was not updated to `8.0.0` but running `uv sync` will update this to current pinecone python sdk version.

.github/workflows/publish-to-pypi.yaml
```diff
@@ -83,7 +83,7 @@ jobs:
 
       - name: Bump pyproject.toml version
         run: |
-          python -c "import re; content = open('pyproject.toml').read(); content = re.sub(r'version = \"[^\"]+\"', 'version = \"${{ steps.bump.outputs.version }}\"', content); open('pyproject.toml', 'w').write(content)"
+          python -c "import re; content = open('pyproject.toml').read(); content = re.sub(r'^version = \"[^\"]+\"', 'version = \"${{ steps.bump.outputs.version }}\"', content, flags=re.MULTILINE); open('pyproject.toml', 'w').write(content)"
```
rohanshah18, line 86: Realized the `target-version` in [pyproject.toml](https://github.com/pinecone-io/pinecone-python-client/pull/542/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L145) was incorrectly changed by the [publish-to-pypi](https://github.com/pinecone-io/pinecone-python-client/pull/542/changes) workflow. The regex re.sub(r'version = \"[^\"]+\"', ...) is too broad and matches both:
version = "7.3.0" (line# 3)
target-version = "py310" (line#125)
When v8.0.0 was released (the first release after the uv migration), both lines were changed to "8.0.0" ([PR](https://github.com/pinecone-io/pinecone-python-client/blob/main/pyproject.toml#L125)). This went unnoticed because the release commit used [skip ci].

